PR #25, opened 2021-02-04
Inmate edit

### LINKED ISSUE #9: [Inmate] DOCs can view and edit inmate permissions
## Motivation
There's currently no way to change to view and edit inmate info

## Task
- [ ] Update UI to be similar to that of staff and request connections (using the antd Table)
- [ ] DOCs can click on the edit button to change inmate information (assign to a different node, change the quota, edit inmate info, etc)

lspdrz:
DO NOT MERGE BEFORE https://github.com/Ameelio/connect-api/pull/154

## Proposed changes

Adds an editable table of the inmates that will update the DB when changed.

## Related Issues

Closes #9 

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

### DIFF
--- a/src/components/editable-table/EditableCell.tsx
+++ b/src/components/editable-table/EditableCell.tsx
@@ -1,0 +1,48 @@
+import { Form, Input, InputNumber } from "antd";
+import React from "react";
+
+interface Props {
+  editing: boolean;
+  dataIndex: string;
+  title: string;
+  inputType: string;
+  record: any;
+  index: number;
+}
+
+const EditableCell: React.FC<Props> = ({
+  children,
+  editing,
+  dataIndex,
+  title,
+  inputType,
+  record,
+  index,
+  ...restProps
+}) => {
+  const inputNode = inputType === "number" ? <InputNumber /> : <Input />;
+  return (
+    <td {...restProps}>
+      {editing ? (
+        <Form.Item
+          name={dataIndex}
+          style={{
+            margin: 0,
+          }}
+          rules={[
+            {
+              required: true,
+              message: `Please Input ${title}!`,
+            },
+          ]}
+        >
+          {inputNode}
+        </Form.Item>
+      ) : (
+        children
+      )}
+    </td>
+  );
+};
+
+export default EditableCell;
--- a/src/components/editable-table/EditableCell.tsx
+++ b/src/components/editable-table/EditableCell.tsx
@@ -5,7 +5,7 @@ interface Props {
   editing: boolean;
   dataIndex: string;
   title: string;
-  inputType: string;
+  inputType: "text" | "number";
   record: any;
   index: number;
 }
--- a/src/components/editable-table/EditableCell.tsx
+++ b/src/components/editable-table/EditableCell.tsx
@@ -31,7 +31,7 @@ const EditableCell: React.FC<Props> = ({
           }}
           rules={[
             {
-              required: true,
+              required: false,
               message: `Please Input ${title}!`,
             },
           ]}
--- a/src/components/editable-table/EditableTable.tsx
+++ b/src/components/editable-table/EditableTable.tsx
@@ -1,12 +1,12 @@
-import { Form, Popconfirm, Table, Typography } from "antd";
+import { Form, Popconfirm, Space, Table, Typography } from "antd";
 import React from "react";
 import { useState } from "react";
-import { Column } from "src/typings/Common";
+import { TableColumn } from "src/typings/Common";
 import EditableCell from "./EditableCell";
 
 interface Props {
-  originalData: any;
-  columns: Column[];
+  originalData: Record<string, any>[];
+  columns: TableColumn[];
   onSave: Function;
 }
 
@@ -61,19 +61,14 @@ export default function EditableTable({
       render: (_: any, record: any) => {
         const editable = isEditing(record);
         return editable ? (
-          <span>
-            <Typography.Link
-              onClick={() => save(record.id)}
-              style={{
-                marginRight: 8,
-              }}
-            >
+          <Space>
+            <Typography.Link onClick={() => save(record.id)}>
               Save
             </Typography.Link>
             <Popconfirm title="Sure to cancel?" onConfirm={cancel}>
               <Typography.Link>Cancel</Typography.Link>
             </Popconfirm>
-          </span>
+          </Space>
         ) : (
           <Typography.Link
             disabled={editingId !== null}
--- a/src/components/editable-table/EditableTable.tsx
+++ b/src/components/editable-table/EditableTable.tsx
@@ -1,0 +1,118 @@
+import { Form, Popconfirm, Table, Typography } from "antd";
+import React from "react";
+import { useState } from "react";
+import EditableCell from "./EditableCell";
+
+interface Props {
+  originalData: any;
+  columns: any[];
+}
+
+const EditableTable = ({ originalData, columns }: Props) => {
+  const [form] = Form.useForm();
+  const [data, setData] = useState(originalData);
+  const [editingKey, setEditingKey] = useState(null);
+
+  const isEditing = (record: any) => record.id === editingKey;
+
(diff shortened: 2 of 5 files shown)


src/pages/Inmate/index.tsx
```diff
@@ -1,19 +1,75 @@
...
+
+const connector = connect(mapStateToProps, mapDispatchToProps);
+
+type PropsFromRedux = ConnectedProps<typeof connector>;
+
+const InmateContainer: React.FC<PropsFromRedux> = ({
+  updateInmate,
+  inmates,
+}) => {
+  const columns = [
```
gabrielsaruhashi, line 26: nit: think we should create a type for this object in Common.ts to avoid using any

src/redux/modules/inmate.ts
```diff
@@ -41,4 +41,26 @@ export const inmatesSlice = createSlice({
   },
 });
 
+export const updateInmate = createAsyncThunk(
+  "inmate/updateInmate",
+  async (args: { inmate: Inmate }) => {
```
gabrielsaruhashi, line 46: I _think_ you can just pass inmate directly here
No need to wrap it as an args as it's a single object

src/redux/modules/inmate.ts
```diff
@@ -41,4 +41,26 @@ export const inmatesSlice = createSlice({
   },
 });
 
+export const updateInmate = createAsyncThunk(
```
gabrielsaruhashi, line 44: you need to create a reducer to update the store after this change happens
check the extraReducers below for an example

updateStaff is a good example

src/components/editable-table/EditableTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+      title: "",
+      dataIndex: "operation",
+      render: (_: any, record: any) => {
+        const editable = isEditing(record);
+        return editable ? (
+          <span>
+            <Typography.Link
+              onClick={() => save(record.id)}
+              style={{
+                marginRight: 8,
```
gabrielsaruhashi, line 63: do we need this margin here?

src/components/editable-table/EditableTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { useState } from "react";
+import EditableCell from "./EditableCell";
+
+interface Props {
+  originalData: any;
+  columns: any[];
+  onSave: Function;
+}
+
+const EditableTable = ({ originalData, columns, onSave }: Props) => {
```
gabrielsaruhashi, line 12: nit: instead of using const

could we just do a one-line 

export default function EditableTable?

I've been using the one line function export for components latley

src/components/editable-table/EditableTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+        newData.splice(index, 1, { ...item, ...row });
+        await onSave(newData[index]);
+        setData(newData);
+        setEditingId(null);
+      } else {
+        newData.push(row);
+        setData(newData);
+        setEditingId(null);
+      }
+    } catch (errInfo) {
```
gabrielsaruhashi, line 47: nit; I've been using err as the variable name for catch clauses

src/components/editable-table/EditableTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+            onClick={() => edit(record)}
+          >
+            Edit
+          </Typography.Link>
+        );
+      },
+    },
+  ]);
+  const mergedColumns = columnsWithEdit.map((col) => {
+    if (!col.editable) {
```
gabrielsaruhashi, line 84: nit: to make it short and sweet, I think we could just turn this into a one line ternary operator return

src/components/editable-table/EditableTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <Table
+        components={{
+          body: {
+            cell: EditableCell,
+          },
+        }}
+        bordered
+        dataSource={data}
+        columns={mergedColumns}
+        rowClassName="editable-row"
```
gabrielsaruhashi, line 110: is this class being used?

src/components/editable-table/EditableCell.tsx
```diff
@@ -0,0 +1,48 @@
+import { Form, Input, InputNumber } from "antd";
+import React from "react";
+
+interface Props {
+  editing: boolean;
+  dataIndex: string;
+  title: string;
+  inputType: string;
```
gabrielsaruhashi, line 8: it seems like you're only using "text" and "number"
can we enforce these types for better readability?

src/components/editable-table/EditableTable.tsx
```diff
@@ -0,0 +1,119 @@
+import { Form, Popconfirm, Table, Typography } from "antd";
+import React from "react";
+import { useState } from "react";
+import EditableCell from "./EditableCell";
+
+interface Props {
+  originalData: any;
+  columns: any[];
```
gabrielsaruhashi, line 8: see comment below re: avoiding use of any here

src/components/editable-table/EditableTable.tsx
```diff
@@ -0,0 +1,119 @@
+import { Form, Popconfirm, Table, Typography } from "antd";
+import React from "react";
+import { useState } from "react";
+import EditableCell from "./EditableCell";
+
+interface Props {
+  originalData: any;
```
gabrielsaruhashi, line 7: isnt originalData always an array?

think it makes sense to use Object[]

gabrielsaruhashi (changes requested):
congrats on your first client PR 🎉 
love the edit table abstraction

src/components/editable-table/EditableTable.tsx
lspdrz: Yep, it adds some space between the "Save" and "Cancel" buttons

src/components/editable-table/EditableTable.tsx
lspdrz: No, removed

src/components/editable-table/EditableTable.tsx
lspdrz: Done

src/components/editable-table/EditableTable.tsx
lspdrz: If we use Object instead, TS complains about fields from the array we use later on (id)

src/pages/Inmate/index.tsx
gabrielsaruhashi: nit: wanna type it with the type you created?

src/pages/Inmate/index.tsx
```diff
@@ -1,19 +1,75 @@
...
+  inmates: selectAllInmates(state),
+});
+
+const mapDispatchToProps = { updateInmate };
+
+const connector = connect(mapStateToProps, mapDispatchToProps);
+
+type PropsFromRedux = ConnectedProps<typeof connector>;
+
+const InmateContainer: React.FC<PropsFromRedux> = ({
```
gabrielsaruhashi, line 22: nit: can we call this UnconnectedInmateContainer

src/redux/modules/inmate.ts
```diff
@@ -38,6 +67,16 @@ export const inmatesSlice = createSlice({
     builder.addCase(fetchInmates.rejected, (state, action) =>
       console.log("error")
     );
+    builder.addCase(updateInmate.fulfilled, (state, action) =>
```
gabrielsaruhashi, line 70: awesome! tks

gabrielsaruhashi (approved):
🚀

PR #64, opened 2021-04-11
pending calls

gabrielsaruhashi:
## Description
- Refactored the requests page to follow the Next.js convention proposed by @lspdrz 
- DOCs can see page with two tabs (connection requests and call requests0
- Can approve/reject
- Better proxy setup to avoid Cors issues in staging server

### DIFF
--- a/src/components/Requests/Requests.tsx
+++ b/src/components/Requests/Requests.tsx
@@ -1,0 +1,61 @@
+import React, { useState } from "react";
+import { Table, Space, Layout, Avatar, Button, Tabs } from "antd";
+import Header from "src/components/Header/Header";
+import { Connection } from "src/typings/Connection";
+import { Call } from "src/typings/Call";
+import { Inmate } from "src/typings/Inmate";
+import { WRAPPER_STYLE } from "src/styles/styles";
+import ConnectionRequests from "./ConnectionRequests";
+import { CallRequests } from "./CallRequests";
+
+const { Column } = Table;
+const { Content } = Layout;
+
+interface Props {
+  connections: Connection[];
+  acceptConnection: (connection: Connection) => void;
+  rejectConnection: (connection: Connection) => void;
+  calls: Call[];
+  acceptCall: (call: Call) => void;
+  rejectCall: (call: Call) => void;
+}
+
+type TTab = "connections" | "calls";
+
+const Requests = ({
+  connections,
+  acceptConnection,
+  rejectConnection,
+  calls,
+  acceptCall,
+  rejectCall,
+}: Props) => {
+  const [activeTab, setActiveTab] = useState<TTab>("connections");
+
+  return (
+    <Content>
+      <Header
+        title="Approval Requests"
+        subtitle="Review all connection requests between incarcerated people in your facility and their loved one on the outside."
+      />
+      <Tabs
+        style={WRAPPER_STYLE}
+        defaultActiveKey={activeTab}
+        onChange={(tab: string) => setActiveTab(tab as TTab)}
+      >
+        <Tabs.TabPane tab="Connections" key="connections">
+          <ConnectionRequests
+            connections={connections}
+            accept={acceptConnection}
+            reject={rejectConnection}
+          />
+        </Tabs.TabPane>
+        <Tabs.TabPane tab="Calls" key="calls">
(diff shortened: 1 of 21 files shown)


src/hooks/useCalls.ts
```diff
@@ -24,3 +24,22 @@ export function useCalls() {
...
+    const pendingCalls = baseCalls.filter(
+      (call) => call.status === "pending_approval"
+    );
+    setCalls(
+      loadAllCallEntities(pendingCalls, contactEnts, inmateEnts, kioskEnts)
+    );
+  }, [baseCalls, inmateEnts, contactEnts, kioskEnts]);
+
+  return calls;
+}
```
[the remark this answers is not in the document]
mpekala23: just to clarify though, the benefit to using this over a more custom selector is that you can automatically make the api call here when any of the entities change?

src/hooks/useConnections.ts
```diff
@@ -0,0 +1,62 @@
...
+  const connections = useConnections();
+
+  useEffect(() => {
+    setRequests(
+      connections.filter((connection) => connection.status === "pending")
+    );
+  }, [connections]);
+
+  return requests;
+}
```
mpekala23, line 41: i'm definitely fine with using hooks, but could this also be done using a selector?

src/redux/modules/call.ts
```diff
@@ -72,6 +97,9 @@ export const callsSlice = createSlice({
     builder.addCase(fetchCallMessages.fulfilled, (state, action) =>
       callsAdapter.updateOne(state, action.payload)
     );
+    builder.addCase(updateCallStatus.fulfilled, (state, action) =>
+      callsAdapter.updateOne(state, action.payload)
+    );
```
mpekala23, line 102: should there also be pending and rejected cases for these?

src/redux/selectors.ts
```diff
@@ -50,40 +50,12 @@ export const {
...
-  if (!inmate) return;
-  if (!contact) return;
-  return { inmate, contact, ...connection };
-};
-
-export const selectConnectionRequests = (state: RootState) => {
-  return selectAllConnections(state).filter(
-    (connection) => connection.status === "pending"
-  );
-};
```
mpekala23, line 70: so these are being deleted in favor of the hooks? is there a performance benefit? again, I'm don't really have a strong opinon just curious what the motiviation is

src/typings/Call.ts
```diff
@@ -65,7 +65,7 @@ export type CallStatus =
   | "terminated"
   | "no_show";
 
-type ISOString = string;
+export type ISOString = string;
```
mpekala23, line 68: is this just for type clarity?

mpekala23 (approved):
just a couple questions related to using custom hook vs custom selector, good either way tho

src/hooks/useConnections.ts
gabrielsaruhashi: this one ye!!!
My assumption is that it's better to do either one or the other to keep it consistent

src/redux/modules/call.ts
gabrielsaruhashi: ye!! i'll do error handling later :)

src/typings/Call.ts
gabrielsaruhashi: yep! i've realized these help a lot while working on the CVH Handler

It will help for readability when new devs onboard.

Seeing a plain "string" for a date field is kinda bizarre

src/redux/selectors.ts
gabrielsaruhashi: ye! my problem was the normalization step
bc that requires access to the state of the store

src/components/Requests/Requests.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface Props {
+  connections: Connection[];
+  acceptConnection: (connection: Connection) => void;
+  rejectConnection: (connection: Connection) => void;
+  calls: Call[];
+  acceptCall: (call: Call) => void;
+  rejectCall: (call: Call) => void;
+}
+
+type TTab = "connections" | "calls";
```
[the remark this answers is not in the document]
gabrielsaruhashi: Just a auxiliary type for the tab keys!

PR #75, opened 2021-05-20
show private files on contact page

lspdrz:
Private file viewing for DOC client

### DIFF
--- a/src/pages/Contact/index.tsx
+++ b/src/pages/Contact/index.tsx
@@ -50,6 +50,7 @@ function ContactPage({
     selectContactById(state, match.params.id)
   );
 
+  // TODO: Implement a way to cache these files: https://github.com/Ameelio/connect-doc-client/issues/76
   useEffect(() => {
     const getContactImages = async () => {
       if (contact) {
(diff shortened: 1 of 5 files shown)


src/pages/Contact/index.tsx
```diff
@@ -42,6 +49,19 @@ function ContactPage({
   const contact = useSelector((state: RootState) =>
     selectContactById(state, match.params.id)
   );
+
+  useEffect(() => {
+    const getContactImages = async () => {
```
gabrielsaruhashi, line 54: a downside to this approach is that we don't store previous API requests for IDs, meaning it'll re-fetch every time even if the user had already loaded this page once.

It's fine for now , but I'll open an issue as an enhancement for the future.

src/typings/IdentificationImage.ts
```diff
@@ -0,0 +1,16 @@
+interface IdentificationImage {
+  id: string;
+  createdAt: string;
+  updatedAt: string;
+  name: string;
+  ownerMalanId: string;
+  key: string;
+  category: string;
```
gabrielsaruhashi, line 8: shouldnt category be typed?

gabrielsaruhashi (approved):
approved with a couple of comments!

src/pages/Contact/index.tsx
gabrielsaruhashi: @lspdrz could you just write a TODO comment and link the issue?
https://github.com/Ameelio/connect-doc-client/issues/76

PR #94, opened 2021-07-12
DOC stays connected to the call globally

gabrielsaruhashi:
## Motivation
- Before, DOC could not navigate to other tabs within Ameelio Connect w/o the calls dropping.
- The live calls page was very vulnerable to unmounts 

## Solution
Moving the room client and socket logic to a separate higher level hook that can be accessed by the React Context API.

Because the socket connections are only used on this one page, it didn't make sense to me to add it to the Redux store. Also Redux would be unhappy about storing the RoomClient class.

After doing some research on different approaches to web socket connection, I thought the Context API was modular and easy to implement.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -31,6 +31,8 @@ import { useCallCountWithStatus, useCallsWithStatus } from "./hooks/useCalls";
 import Modals from "./components/Modals";
 import * as Sentry from "@sentry/react";
 import { Integrations } from "@sentry/tracing";
+import { LiveCallsContext } from "./context/LiveCallsContext";
+import { useLiveCalls } from "./hooks/useLiveCalls";
 
 const LOGIN_PATH = "/login";
 
@@ -110,6 +112,8 @@ function App({ history }: { history: History }) {
     }
   }, [selected, dispatch]);
 
+  const liveCallsContextValue = useLiveCalls();
+
   return (
     <ConnectedRouter history={history}>
       <Layout style={{ minHeight: "100vh" }}>
@@ -133,18 +137,20 @@ function App({ history }: { history: History }) {
           session.status === "loading" ? (
             <Loader />
           ) : (
-            <Switch>
-              <Route exact path={LOGIN_PATH} component={Login}></Route>
-              {ROUTES.map((route) => (
-                <ProtectedRoute
-                  exact
-                  {...defaultProtectedRouteProps}
-                  path={route.path}
-                  component={route.component}
-                  key={route.label}
-                ></ProtectedRoute>
-              ))}
-            </Switch>
+            <LiveCallsContext.Provider value={liveCallsContextValue}>
+              <Switch>
+                <Route exact path={LOGIN_PATH} component={Login}></Route>
+                {ROUTES.map((route) => (
+                  <ProtectedRoute
+                    exact
+                    {...defaultProtectedRouteProps}
+                    path={route.path}
+                    component={route.component}
+                    key={route.label}
+                  ></ProtectedRoute>
+                ))}
+              </Switch>
+            </LiveCallsContext.Provider>
           )}
         </Layout>
       </Layout>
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -133,18 +137,20 @@ function App({ history }: { history: History }) {
...
-                <ProtectedRoute
-                  exact
-                  {...defaultProtectedRouteProps}
-                  path={route.path}
-                  component={route.component}
-                  key={route.label}
-                ></ProtectedRoute>
-              ))}
-            </Switch>
+            <LiveCallsContext.Provider value={liveCallsContextValue}>
```
mpekala23, line 140: fancy fancy

src/hooks/useLiveCalls.ts
```diff
@@ -0,0 +1,213 @@
...
+        ) => {
+          console.log(
+            `[VideoChat] Received ${kind} consume from ${participant.type} ${participant.id}`
+          );
+          if (kind === "video") {
+            setRemoteVideos((remotes) => ({
+              ...remotes,
+              [callId]: {
+                ...remotes[callId],
+                [getParticipantStreamId(participant, callId)]: stream,
```
mpekala23, line 73: out of curiosity, what does the indexing within `[ ]` do here? like how does `[callId]:` differ from `callId:`

src/hooks/useLiveCalls.ts
```diff
@@ -0,0 +1,213 @@
...
+
+            return { ...otherRemotes, [callId]: otherStreams };
+          });
+        }
+      );
+
+      // listening for text messages
+      rc.socket.on("callStatus", async (status: InCallStatus) => {
+        console.log("[VideoChat] Received status update", status);
+        // TODO: check this works
```
mpekala23, line 119: did you do this?

src/hooks/useLiveCalls.ts
```diff
@@ -0,0 +1,213 @@
...
+      // initialize new sockets once
+      if (target in temp) continue;
+
+      const newSocketClient = io.connect(target, { transports: ["websocket"] });
+      temp[target] = newSocketClient;
+      joinRoom(call.id, newSocketClient);
+    }
+    console.log("[Fresh calls] New socket clients", temp);
+    setSocketMap((curr) => ({ ...curr, ...temp }));
+  }, [freshCalls, joinRoom]);
```
mpekala23, line 188: very very cool & clever

src/hooks/useLiveCalls.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      for (const [callId, rc] of Object.entries(roomClients)) {
+        rc.socket.emit("heartbeat", { callId });
+      }
+    }, 1000);
+    return () => clearInterval(interval);
+  }, [roomClients]);
```
mpekala23, line 197: oh so this is where this comes from. cool cool

src/hooks/useLiveCalls.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  // useEffect(() => {
+  //   return () => {
+  //     for (const rc of Object.values(roomClients)) {
+  //       console.log("[Cleanup] Destroying room clients");
+  //       rc.destroy();
+  //     }
+  //   };
+  // }, [roomClients]);
```
mpekala23, line 206: is it safe to have this commented out?

src/typings/Common.ts
```diff
@@ -25,3 +25,5 @@ export interface RejectionReason {
   description: string;
   key: string;
 }
+
+export type StreamId = string;
```
mpekala23, line 29: is this worth having?

mpekala23 (commented):
very late but very cool! contexts definitely seems cleaner than the ref / redux solution on mobile. probably won't try and tackle the migration there until something drastic needs to change (multiple people on a call perhaps)

src/hooks/useLiveCalls.ts
gabrielsaruhashi: so [callId] is a computed value
callId is considered to be a string literal 'callId'

src/hooks/useLiveCalls.ts
gabrielsaruhashi: ya! oop will dellete this

src/typings/Common.ts
gabrielsaruhashi: think it makes it more descriptive and readable! String is a bit hard to understand

PR #95, opened 2021-07-22
Jason's requests for ICIW's facility-wide launch tomorrow

gabrielsaruhashi:
## Proposed Changes
- Can download filtered logs on 'Search for Visits' page
- Can view kiosk name assigned to the call on 'Search for Visits'
- Initial view of 'Search for Visits' only surfaces upcoming calls


src/components/Dashboard/DailyReport.tsx
```diff
@@ -26,29 +26,41 @@ const styles = StyleSheet.create({
...
-          )}`}
-          subtitle={`${facility.name}`}
-        />
+        <Header title={filename} subtitle={facilityName} />
         {Object.keys(callBlocks).map((block) => (
           <View key={block} style={styles.blockContainer}>
-            <SectionHeader title={format(new Date(block), "HH:mm aaa")} />
+            <SectionHeader
+              title={format(new Date(block), "HH:mm aaa, M/d/yy")}
+            />
```
mpekala23, line 63: not sure if this is important but isn't this not guaranteed to be sorted? might it be a clearer doc if you sorted these start times keys before displaying them?

src/components/DownloadScheduleMenu/DownloadScheduleMenu.tsx
```diff
@@ -0,0 +1,58 @@
...
+        />
+      </Menu.Item>
+      <Menu.Item key="2" icon={<FilePdfOutlined />}>
+        <PDFDownloadButton
+          calls={calls}
+          facilityName={facilityName}
+          canViewDetails={false}
+          filename={filename}
+        />
+      </Menu.Item>
```
mpekala23, line 37: what exactly is the difference between these two? it looks like they have the same icon, only differing in the `canViewDetails` prop?

mpekala23 (approved):
looks good to me! a couple questions!

src/components/Dashboard/DailyReport.tsx
gabrielsaruhashi: yep!! guaranteed to be sorted (at least for now).
The API sorts by scheduled sstart

src/components/DownloadScheduleMenu/DownloadScheduleMenu.tsx
gabrielsaruhashi: yup! One has private information about the visitors, one does not

There are two copies of this daily report. One used internally by the DOC, one that can be printed and hung on the wall of the inc people bulletin board

PR #100, opened 2021-08-09
pdf wont get cut off

gabrielsaruhashi:
## Proposed changes
Prevents new call blocks from getting wrapped/cut off in between pages.

## Types of changes

What types of changes does your code introduce to Appium?
_Put an `x` in the boxes that apply_

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)


src/components/Dashboard/DailyReport.tsx
```diff
@@ -16,29 +16,22 @@ const styles = StyleSheet.create({
...
 
 const DailyReport: React.FC<Props> = ({
   calls,
   facilityName,
-  canViewDetails,
+  showContactDetails,
   filename,
 }) => {
-  const [callBlocks, setCallBlocks] = useState<Record<string, Call[]>>({});
+  const [callBlocks, setCallBlocks] = useState<Record<ISOString, Call[]>>({});
```
cainlevy, line 34: TypeScript implements structural typing, not nominal, so `ISOString` and `string` are [equivalent in every way](https://www.typescriptlang.org/play?#code/C4TwDgpgBAkgygeTsATgSwHYHMoF4oDOqmWA3AFADGA9hkVALYjxLHZ5QDkG1KDAhgBtCbLJyj8CI9Ngo06wRiGQyc+brwHCiq8ZNiIVJUkA) except as documentation to humans.

As a human, I'm curious if there's anything here I should know about ISO strings?

src/components/Dashboard/PDFDownloadButton.tsx
```diff
@@ -1,20 +1,20 @@
 import React, { useEffect, useState } from "react";
 import { Spin, Typography } from "antd";
-import { PDFDownloadLink } from "@react-pdf/renderer";
+import { PDFDownloadLink, usePDF } from "@react-pdf/renderer";
```
cainlevy, line 3: needed?

src/components/Dashboard/DailyReportCallBlock.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+const styles = StyleSheet.create({
+  page: {
+    padding: 30,
+  },
+  container: {
+    flex: 1,
+    flexDirection: "row",
+  },
```
cainlevy, line 24: :scissors:

cainlevy (approved):
how does the pdf generator decide where to insert page breaks? will it be smart enough to place them between call blocks when that gap is available, or did i miss some hint we provide to tell it where page breaks should go?

src/components/Dashboard/DailyReport.tsx
gabrielsaruhashi: Nope! 
I created that type just for readability / documentation tbh
I thought it could be confusing to just have "string" in there 
Let me know if you think this is an anti-pattern.

gabrielsaruhashi:
@cainlevy 
the combination of wrap={false} + the chunks fo 3 are the heuristic i'm using to make it work! Can DM you a PDF of the final result
It was the quickest way I could think of to solve the problem, but it does mean it ends up having a lot of whitespace

PR #106, opened 2021-09-09
getting started

cainlevy:
Accumulated changes from installing connect-doc-client.

Note that on MacOS there is still an issue with the `avatar` and `pdf` folders, which have both uppercase and lowercase versions. The git checkout collapses them because of MacOS' case insensitive file system, but import paths may reference the wrong folder.


package.json
```diff
@@ -33,7 +22,6 @@
     "http-proxy-middleware": "^1.1.1",
     "lodash": "^4.17.20",
     "mediasoup-client": "^3.6.23",
-    "node-sass": "^4.14.1",
```
cainlevy, line 36: nothing used it! 🎉

lspdrz (approved):
LGTM

PR #111, opened 2021-10-05
Revamped download schedule menu

gabrielsaruhashi:
## Proposed changes
**PDF Enabled**
image

**PDF Disabled**
image


src/redux/modules/staff.ts
```diff
@@ -67,7 +67,7 @@ export const staffSlice = createSlice({
   },
   extraReducers: (builder) => {
     builder.addCase(fetchStaff.fulfilled, (state, action) =>
-      staffAdapter.addMany(state, action.payload)
+      staffAdapter.setAll(state, action.payload)
```
gabrielsaruhashi, line 70: ensures we don't load DOC staff from other facilities when we switch faciltiies

PR #112, opened 2021-10-05
DOC can edit the connection directly from the visitor/inc person's profile

gabrielsaruhashi:
## Proposed changes
image


src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -1,40 +1,83 @@
...
         <Space direction="vertical">
           <Avatar fallback={name} size={48} />
           <Typography.Text>{name}</Typography.Text>
-          <Typography.Text type="secondary">{relationship}</Typography.Text>
+          <Typography.Text type="secondary">
+            {connection.relationship}
+          </Typography.Text>
         </Space>
       </div>
+      <Modal
```
gabrielsaruhashi, line 53: If we ever see a use case for reusing this modal, I'd move it to the modals slice
For now, it feels really nice to reduce complexity and keep it close to the only component that uses it.

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -1,40 +1,83 @@
...
+  const [open, setOpen] = useState(false);
+  const person = type === "inmate" ? connection.contact : connection.inmate;
   const name = genFullName(person);
+
+  const renderButton = (): JSX.Element => {
+    switch (connection.status) {
+      case "active":
+        return (
+          <Button onClick={() => onUpdate("rejected")} danger type="primary">
+            Disable
```
gabrielsaruhashi, line 28: I scoped out a "are you sure you want to do this action" flow because this action is low-stakes, and pretty easily reversible. 

The doc dash issues a notif as well to acknowledge the action

cainlevy: does "scoped out" mean "in scope" or "not in scope"?

are notifications sent to users?

src/components/Profile/Connections/ConnectionItem.tsx
cainlevy: "modals slice" 😲 

what hath redux wrought.

> For now, it feels really nice to reduce complexity and keep it close to the only component that uses it.

sgtm!

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -1,40 +1,83 @@
...
-  relationship,
-  type,
-  navigate,
-}) => {
+const ConnectionItem: React.FC<Props> = ({ connection, type, onUpdate }) => {
+  const [open, setOpen] = useState(false);
+  const person = type === "inmate" ? connection.contact : connection.inmate;
   const name = genFullName(person);
+
+  const renderButton = (): JSX.Element => {
```
cainlevy, line 23: this is just another react component. let's name and render it as such.

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -1,40 +1,83 @@
...
+              <ContactCell contact={connection.contact} />
+            </Descriptions.Item>
+            <Descriptions.Item label="Created At">
+              {formatDate(new Date(connection.createdAt), "date")}
+            </Descriptions.Item>
+            <Descriptions.Item label="Status">
+              <ConnectionTag status={connection.status} />
+            </Descriptions.Item>
+          </Descriptions>
+          <div className="ml-auto">{renderButton()}</div>
```
cainlevy, line 78:
```suggestion
          <div className="ml-auto"><MyButton /></div>
```

src/components/Profile/Connections/ConnectionTag.tsx
```diff
@@ -0,0 +1,27 @@
...
+      case "active":
+        return "green";
+      case "rejected":
+        return "red";
+      case "inactive":
+        return "default";
+      case "pending":
+        return "yellow";
+    }
+  };
```
cainlevy, line 22: let's hoist this out. it can take `status` as an arg instead of relying on a closure.

another trick is to turn this into simple configuration data with typescript guarantees:

```
const StatusColors: Record<ConnectionStatus, PresetColorType | PresetStatusColorType> = {
  ...
}
```

src/utils/Time.ts
```diff
@@ -0,0 +1,12 @@
...
+export function formatDate(
+  date: Date,
+  type: "date" | "time" | "datetime" | "dateyear"
+): string {
+  if (type === "date") return format(date, "MM/dd/yy");
+  if (type === "time") return format(date, "HH:mm");
+  if (type === "datetime") return format(date, "MM/dd/yy HH:mm");
+  if (type === "dateyear") return format(date, "yyyy-MM-dd");
+  return "";
+}
```
cainlevy, line 12: this will be really helpful if we need to localize .... wait we don't need to localize. wow i'm still getting used to working on an app that may never go global.

cainlevy (approved):

src/components/Profile/Connections/ConnectionItem.tsx
gabrielsaruhashi: Not in scope for now! 
Adding a ticket and marking it as a future improvement.

gabrielsaruhashi: And yes notifs are sent :)

src/utils/Time.ts
gabrielsaruhashi: who knows ... Ameelio Brazil might be a thing 👀

src/components/Profile/Connections/ConnectionItem.tsx
gabrielsaruhashi: image

sample of notifs created after this DOC action

PR #121, opened 2021-10-15
Fixes infinite loop on Contact page

lspdrz:
There was an infinite loop on a useEffect here that was visible in the API logs as there was a call on each iteration


src/pages/ContactPage.tsx
```diff
@@ -31,7 +31,7 @@ function ContactPage({ match }: RouteComponentProps<TParams>): ReactElement {
   useEffect(() => {
     if (!contact) return;
     dispatch(fetchContactIdImages(contact.id));
-  }, [contact, dispatch]);
+  }, []);
```
cainlevy, line 34: so previously the `fetchContactIdImages` would change `contact` and re-trigger this effect?

does this component need to support switching between contacts while mounted?

gabrielsaruhashi (approved):

PR #125, opened 2021-10-20
adds edit quota feature to inmate profile page

lspdrz:
Note: cannot merge until PR 365 from connect-api-nest has been merged and deployed

[sc-1101]

https://user-images.githubusercontent.com/9705113/138141221-5f508a0a-0d52-4f37-b435-7cab0c195935.mov


src/components/Profile/ProfileHeader/EditQuota.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { InputNumber } from "antd";
+
+interface Props {
+  quota: number;
+  changeQuota: (value: number) => void;
```
gabrielsaruhashi, line 6: nit: I'd rename it to "onChange"
It's the new convention we're using based on Lance's feedback

src/components/Profile/ProfileHeader/EditQuota.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+const EditQuota: React.FC<Props> = ({ quota, changeQuota }) => {
```
gabrielsaruhashi, line 9: Could we do
const EditQuota = ({quota, changeQuota}: Props) => {
}

for consistency? That's what we do in most plcaces

src/components/Profile/ProfileHeader/EditQuota.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <>
+      <InputNumber
+        min={0}
+        max={100}
```
gabrielsaruhashi, line 14: 100 seems high? Bring it down to 15?

src/redux/modules/inmate.ts
```diff
@@ -31,6 +31,25 @@ export const fetchInmateAvailability = createAsyncThunk<
   return (body.data as RawOpenings[]).map((raw) => cleanSlots(raw)).flat();
 });
 
+export const updateInmate = createAsyncThunk(
```
gabrielsaruhashi, line 34: could you type the thunk?

here's an example on how to do it. pulled it from the inmate module
image

src/components/Profile/ProfileHeader/EditQuota.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+interface Props {
+  quota: number;
+  changeQuota: (value: number) => void;
+}
+
+const EditQuota: React.FC<Props> = ({ quota, changeQuota }) => {
+  return (
+    <>
+      <InputNumber
```
gabrielsaruhashi, line 12: is this child component helping us? I don't know if it's doing much work at all

src/components/Profile/ProfileHeader/InmateProfileHeader.tsx
```diff
@@ -15,6 +19,22 @@ const InmateProfileHeader = ({
   facilityName,
   onScheduleClick,
 }: Props) => {
+  const dispatch = useAppDispatch();
+  let [quotas, setQuotas] = useState({
```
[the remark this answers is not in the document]
gabrielsaruhashi: what's the benefit of having these two nested under the same object, rather than tracking them separately?

They don't necessarily get updated at the same time. This gives the impression that they are supposed to.

I.e I'd move them to

const [callQuota, setCallQuota]
const [visitQuota, setVisitQuota]

gabrielsaruhashi (changes requested):
last round of feedback

src/redux/modules/inmate.ts
```diff
@@ -44,6 +59,14 @@ export const inmatesSlice = createSlice({
     builder.addCase(fetchInmates.rejected, (state, action) =>
       console.log("error")
     );
+    builder.addCase(updateInmate.fulfilled, (state, action) => {
```
gabrielsaruhashi, line 62: could u please add a case for "rejection", and showToast describing the error.
I've done in a couple of other reducers if u want an example

gabrielsaruhashi (approved):
take a look at the two final comments?

PR #138, opened 2021-10-29
Add CD config for prod deploys

FreedomBen:
These are triggered by a tag

Also add slack notifications for deploys


.github/workflows/build-test-deploy.yml
```diff
@@ -3,16 +3,15 @@ on:
   push:
     branches:
       - main
-      - dev
-      - add-ci-cd
+      - add-ci-cd-prod
```
gabrielsaruhashi, line 6: is this for testing?

gabrielsaruhashi: / do we need it merged?

README.md
```diff
@@ -43,39 +43,49 @@ You can learn more in the [Create React App documentation](https://facebook.gith
...
+
+### Triggering a deploy
+
+#### Staging
+
+Staging deploys happen automatically when your PR is merged to master/main!  You don't have to do anything else.
+
+#### Prod
+
+Prod deploys happen when you tag a commit with a tag name starting with 'prod-'.  The 'prod-' prefix is intended to prevent accidental deployments to production where one is not intending it.  To tag the top commit for deployment to production:
```
gabrielsaruhashi, line 70: thank you!
Do we have preference to include everything in a single ReadME vs splitting up into a separate md called Releasing.md?

Asking because we do the latter on other repos

scripts/deploy-release.sh
```diff
@@ -408,6 +411,36 @@ sanitize ()
   echo "${1}" | sed -e 's/./*/g' | awk '{ print substr($0, 0, 10) }'
 }
 
+get_gh_link ()
+{
+  echo "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+}
+
+send_slack_message ()
```
gabrielsaruhashi, line 419: v cool!!!

scripts/deploy-release.sh
```diff
@@ -408,6 +411,36 @@ sanitize ()
...
+}
+
+notify_deploy_start ()
+{
+  send_slack_message ":hourglass:  ${ENV} deploy started for ${GITHUB_REPOSITORY} by $GITHUB_ACTOR.  To see details: $(get_gh_link)"
+}
+
+notify_deploy_fail ()
+{
+  send_slack_message ":x:  ${ENV} deploy for ${GITHUB_REPOSITORY} by $GITHUB_ACTOR has FAILED!.  To see details: $(get_gh_link)"
```
gabrielsaruhashi, line 436: should we also include the build version on the message itself?
If it's extra work, no sweat, was just curious if there's more info we'd like to know

gabrielsaruhashi: ^ this q applies to every notification

gabrielsaruhashi (approved):

README.md
FreedomBen: Oh no I don't have a preference.  After verifying the prod deploy I'll push a final "cleanup" that removes the one-off branch and moves this to a RELEASE.md

scripts/deploy-release.sh
FreedomBen: yeah I like that idea!  Should be super easy to do

PR #139, opened 2021-11-01
creates initial admin page for superusers

lspdrz:
DO NOT MERGE UNTIL THIS IS MERGED AND DEPLOYED: https://github.com/Ameelio/connect-api-nest/pull/412

Creates the initial Admin page that is only accessible by ameelio superusers. Facility and Staff creation tabs have not been made yet.

Demo below:

https://user-images.githubusercontent.com/9705113/139685650-b05e5659-3988-40c1-a491-039434f2a17f.mov


gabrielsaruhashi:
@lspdrz could you include a video or screenshot to the PR?
Think it's a convention we'd like to keep using as much as possible on client-side changes

src/typings/Session.ts
```diff
@@ -3,6 +3,7 @@ import { Staff } from "./Staff";
 export interface User extends BasePersona {
   email: string;
   staffPositions: Staff[];
+  isSuperUser: boolean;
```
gabrielsaruhashi, line 6: wondering if we want to store the "role" string instead of a boolean 
based on our discussion last week, it seemed like we might want to create distiction across super users in the future

src/pages/AdminPage.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const facility = useAppSelector((state) => state.facilities.selected);
+  const session = useAppSelector((state) => state.session);
+  const kiosks = useAppSelector(selectAllKiosks);
+  const dispatch = useAppDispatch();
+
+  // Only to be used by super users
+  if (!session.user.isSuperUser) return <div />;
+  if (!facility) return <div />;
+
+  const handleCallHoursChange = (
```
gabrielsaruhashi, line 22: why is the call hours here duplicated here?
That should be available to DOC admins

lspdrz:
> @lspdrz could you include a video or screenshot to the PR? Think it's a convention we'd like to keep using as much as possible on client-side changes

For sure, added one above

src/typings/Session.ts
lspdrz: I'm passing the entire roles array and determined admin users from that now

src/pages/AdminPage.tsx
lspdrz: I've just copied over the Settings component stuff as a placeholder. I will remove in the create facility page PR. Please disregard for now

src/utils/Common.ts
```diff
@@ -3,6 +3,9 @@ import { addSeconds, format, differenceInSeconds } from "date-fns";
 import { snakeCase } from "snake-case";
 import { toQueryString } from "src/api/Common";
 import { MeetingFilters, Call } from "src/typings/Call";
+import { User } from "src/typings/Session";
+
+export const isSuperUser = (user: User) => user.roles.includes("admin");
```
gabrielsaruhashi, line 8: v handy

src/pages/AdminPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const groupTree = useAppSelector((state) => state.groups.nodes);
+  const groupEnts = useAppSelector(selectGroupEntities);
+  const facility = useAppSelector((state) => state.facilities.selected);
+  const session = useAppSelector((state) => state.session);
+  const kiosks = useAppSelector(selectAllKiosks);
+  const dispatch = useAppDispatch();
+
+  // Only to be used by super users
+  if (!isSuperUser(session.user)) return <div />;
+  if (!facility) return <div />;
```
gabrielsaruhashi, line 21: Instead of returning <div />
Could you return 
a 403 result?

Example here: https://ant.design/components/result/#header

gabrielsaruhashi: image

gabrielsaruhashi (approved):

PR #141, opened 2021-11-03
adds editstaff form

lspdrz:
[sc-1214]

DO NOT MERGE UNTIL https://github.com/Ameelio/connect-api-nest/pull/420 IS MERGED AND DEPLOYED

Very rudimentary staff creation form. Does not allow support for editing current staff members yet but the scaffolding is there. Video example below

https://user-images.githubusercontent.com/9705113/140060578-23e28837-cdc0-4d25-a068-46d18cec6006.mov


src/api/Common.ts
```diff
@@ -60,9 +60,8 @@ export async function fetchAuthenticated(
   }
 
   if (response.status !== 200 && response.status !== 201) {
-    throw new Error(
-      `Unexpected ${nodeResource} error. Code ${response.status}.`
-    );
+    const responseJson = await response.json();
+    throw new Error(`${responseJson.message}`);
```
lspdrz, line 64: @gabrielsaruhashi i'm being a little reckless here maybe? will this change affect anything else?

src/utils/constants/Common.ts
```diff
@@ -1,5 +1,11 @@
 import { CallAlert, GridOption } from "src/typings/Call";
 
+export enum StaffRoles {
```
gabrielsaruhashi, line 3: nit: StaffRole

src/api/Common.ts
gabrielsaruhashi: do we really want to expose the error message to a DOC user?
I'd rather keep it opaque. Dev messages are often confusing

gabrielsaruhashi: Sentry is our best friend to capture the actual error

src/redux/modules/staff.ts
```diff
@@ -98,13 +100,21 @@ export const staffSlice = createSlice({
...
-    builder.addCase(createStaff.rejected, (state, action) => ({
-      ...state,
-      error: action.error.message,
-    }));
+    builder.addCase(createStaff.fulfilled, (state, action) => {
+      showToast(CREATE_STAFF, "Staff member successfully created!", "success");
+      staffAdapter.addOne(state, action.payload);
+    });
+    builder.addCase(createStaff.rejected, (state, action) => {
+      showToast(
```
gabrielsaruhashi, line 108: add sentry capture exception here

src/redux/modules/staff.ts
```diff
@@ -98,13 +100,21 @@ export const staffSlice = createSlice({
...
     }));
-    builder.addCase(createStaff.fulfilled, (state, action) =>
-      staffAdapter.addOne(state, action.payload)
-    );
-    builder.addCase(createStaff.rejected, (state, action) => ({
-      ...state,
-      error: action.error.message,
-    }));
+    builder.addCase(createStaff.fulfilled, (state, action) => {
+      showToast(CREATE_STAFF, "Staff member successfully created!", "success");
```
gabrielsaruhashi, line 104: "${staff.firstName} ${staff.lastName} was added successfully as a staff member of the facility! "

src/redux/modules/staff.ts
```diff
@@ -35,20 +36,21 @@ export const updateStaff = createAsyncThunk(
...
+    try {
+      const body = await fetchAuthenticated(`staffMembers`, {
+        method: "POST",
+        body: JSON.stringify(staffData),
+      });
+      const staff = body.data as Staff;
 
-    return staff;
+      return staff;
+    } catch (error) {
```
gabrielsaruhashi, line 51: this catch is not needed. it seems like we're catching and re-throwing the same error message
builder.addcase error is already handling the rejection

src/typings/Common.ts
```diff
@@ -29,3 +29,26 @@ export type Tab<TParam> = {
...
+  enumValues?: string[];
+};
+
+type StaffPositionData = {
+  facilityId: string;
+  role: StaffRole;
+  isActive: boolean;
+};
+
+export type StaffData = {
```
gabrielsaruhashi, line 47: hm I dont think this belongs to Common.ts

can't we just use the existing Staff type.
If not,  I'd just move it to the component itself. It's not being used anywhere else

src/components/Admin/Edit.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { FormItem } from "src/typings/Common";
+
+interface Props {
+  formItems: FormItem[];
+  onCreate: (values: any[]) => void;
+}
+
+const Edit = ({ formItems, onCreate }: Props) => {
+  const [createForm] = Form.useForm();
+  const [createMode, setCreateMode] = useState(false);
```
gabrielsaruhashi, line 12: nit could we prefix it with is to make it more clear that it's a boolean?

isCreating, setIsCreating

src/components/Admin/Edit.tsx
```diff
@@ -0,0 +1,90 @@
...
+  return (
+    <>
+      <Button className="mb-5" onClick={() => setCreateMode(!createMode)}>
+        Add new
+      </Button>
+      {createMode ? (
+        <Form
+          form={createForm}
+          onFinish={saveEntity}
+          // onFinishFailed={onFinishFailed}
```
gabrielsaruhashi, line 63: delete?

src/components/Admin/Edit.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { useState } from "react";
+import { Button, Checkbox, Form, Input, Select } from "antd";
+import { FormItem } from "src/typings/Common";
+
+interface Props {
+  formItems: FormItem[];
```
gabrielsaruhashi, line 6: image
instead of recreating the type, I am pretty sure we could just use the actual FormItemProps from antd

src/typings/Common.ts
```diff
@@ -29,3 +29,26 @@ export type Tab<TParam> = {
   key: TParam;
   tab: string;
 };
+
+export type FormItem = {
+  label: string;
+  name: string;
+  placeholder: any;
+  required: boolean;
+  enumValues?: string[];
```
gabrielsaruhashi, line 38: having enumvalues in there feels a bit confusing
If we keep going with these types, I'd de-compose it into a union instead of keeping it nullable

src/components/Admin/EditStaff.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { StaffData } from "src/typings/Common";
+import { useSelector } from "react-redux";
+
+const EditStaff = () => {
+  const dispatch = useAppDispatch();
+  const facilityId = useSelector(
+    (state: RootState) => state.facilities.selected?.id
+  );
+
+  const saveStaff = (values: any) => {
```
gabrielsaruhashi, line 16: what would it take for us to type this one?

src/components/Admin/Edit.tsx
```diff
@@ -0,0 +1,90 @@
...
+          ></Checkbox>
+        );
+      default:
+        return <Input placeholder={formItem.placeholder} />;
+    }
+  };
+
+  return (
+    <>
+      <Button className="mb-5" onClick={() => setCreateMode(!createMode)}>
```
gabrielsaruhashi, line 56: hmm I dont think this "createMode" is making the UI better
Let's just keep the form visible at all times?

gabrielsaruhashi (changes requested):
it's cool that we're trying to abstract the form editing, but there are so many 'any's right now spread out. Im not sure if this abstraction is sustainable if we can't have some safeguards?
It's making it harder to read as well. I wonder if a simple form would have been more clear here

src/components/Admin/Edit.tsx
```diff
@@ -0,0 +1,90 @@
...
+            onChange={(e) => setValue(e.target.checked, formItem.name)}
+          ></Checkbox>
+        );
+      default:
+        return <Input placeholder={formItem.placeholder} />;
+    }
+  };
+
+  return (
+    <>
```
gabrielsaruhashi, line 55: Could we put this in a Card?
I'll follow the design we did for the Admin page
The floating fields look confusing

src/api/Common.ts
lspdrz: Well the inspiration was from testing. I spent 30 min to an hour trying to figure out why my form wasn't posting to the API correctly precisely because the error message so opaque. It ended up being a simple type error, I was inputting an invalid email, and this wasn't communicated. Not communicating these type errors will definitely be a pain point for the usage of this page

src/redux/modules/staff.ts
lspdrz: I'm not sure this is desired behavior - do we really want to be logging sentry errors for every user mistake that will cause the validation to fall? E.g. mistakes on emails, phone numbers, urls....

Maybe we need to conditionallly send sentry errors? But then that opens the conversation/tasks of improving our validation error messages from the API

src/redux/modules/staff.ts
lspdrz: We need it to access the payload of the error message I believe, which I use in the showToast. If we remove that future and go for the opaque error message, I can remove this catch too.

```rejectWithValue(value, [meta]): rejectWithValue is a utility function that you can return (or throw) in your action creator to return a rejected response with a defined payload and meta. It will pass whatever value you give it and return it in the payload of the rejected action. If you also pass in a meta, it will be merged with the existing rejectedAction.meta.```

File: src/typings/Common.ts (hunk 6 above)
Comment (lspdrz) [reply src/typings/Common.ts:47]: We cannot, they're fundamentally different. And we're also using it in staff.ts in the redux files. I could move it and export it from the Staff.ts typing?

File: src/components/Admin/Edit.tsx (hunk 9 above)
Comment (lspdrz) [reply src/components/Admin/Edit.tsx:6]: These are different custom types for use with the admin pages

File: src/typings/Common.ts (hunk 10 above)
Comment (lspdrz) [reply src/typings/Common.ts:38]: Ok, let's talk about these types today

File: src/components/Admin/EditStaff.tsx (hunk 11 above)
Comment (lspdrz) [reply src/components/Admin/EditStaff.tsx:16]: I could create a type of accepted saveStaff values

StaffValue = string | boolean | StaffRole.

I think this would sort out the anys everywhere

File: src/components/Admin/Edit.tsx (hunk 12 above)
Comment (lspdrz) [reply src/components/Admin/Edit.tsx:56]: Let's talk. This is done to provide scaffolding for a view where we'll have a table to edit existing rows of data. All of this is design is in line with typical admin UIs

**Comment by lspdrz** [comment]
@gabrielsaruhashi let's talk about the abstraction today. The whole point of the abstraction was to have the Edit Page be a reusable component across all admin pages, to introduce a standardized way of adding data to the backend in line with other admin sites I've used, particularly Django Admin. As you can see on the create facility PR, it's already being used a second time. So the reasoning was DRY and simple and standard input across all data entry

File: src/redux/modules/staff.ts (hunk 5 above)
Comment (gabrielsaruhashi) [reply src/redux/modules/staff.ts:51]: Let's go with opaque messages
we're not reviewing copy for the error messages to make sure they're user friendly.
I'd rather just throw on sentry

File: src/redux/modules/staff.ts (hunk 3 above)
Comment (gabrielsaruhashi) [reply src/redux/modules/staff.ts:112]: I see now disregard the sentry comment

File: src/typings/Common.ts (hunk 10 above)
Comment (gabrielsaruhashi) [reply src/typings/Common.ts:38]: It's the same technique we talked about for ModalData

File: src/typings/Common.ts (hunk 6 above)
Comment (gabrielsaruhashi) [reply src/typings/Common.ts:47]: Staff.ts would make more sense
Can we call it TCreateStaff ? Or anything to make it clear what's the use case

File: src/typings/Common.ts (hunk 10 above)
Comment (gabrielsaruhashi) [reply src/typings/Common.ts:38]: https://github.com/Ameelio/connect-doc-client/blob/573e13d1ce8871e382e476727d6b330645202a7b/src/typings/Modal.ts

File: src/components/Admin/Edit.tsx [hunk 14]
```diff
@@ -0,0 +1,86 @@
...
+import { FormItem } from "src/typings/Common";
+
+interface Props {
+  formItems: FormItem[];
+  onCreate: (values: any[]) => void;
+}
+
+const Edit = ({ formItems, onCreate }: Props) => {
+  const [createForm] = Form.useForm();
+  const [isCreating, setIsCreating] = useState(false);
```
gabrielsaruhashi, line 12: is this being used?

src/components/Admin/Edit.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+  };
+
+  return (
+    <Card>
+      <Button className="mb-5" onClick={() => setIsCreating(!isCreating)}>
+        Add new
+      </Button>
+      {isCreating ? (
+        <Form form={createForm} onFinish={saveEntity}>
```
gabrielsaruhashi, line 60: I'd just expose the Form directy. 
This "add new" button is not really improving the UI imo

src/components/Admin/EditStaff.tsx
gabrielsaruhashi: we should just have values: CreateStaffDto imo
and do the cleaning up on the child component

src/api/Common.ts
gabrielsaruhashi: This is gonna impact every single API call the DOC makes.
We don't want to scare the DOC staff with error messages thrown on the API whose copy we're not reviewing. ('Entity not found', etc.)
Debugging as a developer should be possible using "Network" and "Redux Toolkit".
If what we're solving for is giving feedback to andrew, i'd unwrap the thunks

src/components/Admin/Edit.tsx
gabrielsaruhashi: what will it take to type it? This is what I am trying to solve for.
'Any' shouldn't be the typing here
I'd make it a generic type 
onCreate<T>: (values: T) => void;

src/utils/constants/Common.ts
```diff
@@ -1,5 +1,11 @@
 import { CallAlert, GridOption } from "src/typings/Call";
 
+export enum StaffRole {
```
gabrielsaruhashi, line 3: this value already exists in User.d.ts - I'd just make sure we delete that one over there
User.d.ts is legacy --- it shouldn't be tehre
but I'd create 
a new User.ts and have StaffRole export from there (and delete teh one on User.d.ts)

src/api/Common.ts
gabrielsaruhashi: Could we just pass an optional parameter for now? 
surfaceErrorMessage?: boolean

We should only turn it on for form creation right now as the feedback is actually actionable.

For GET requests / etc, I am not confident yet that our error messages are good.

src/components/Admin/EditStaff.tsx
lspdrz: We can't do that and maintain the abstraction of the child component to be used with different parent components and different form structures

src/utils/constants/Common.ts
lspdrz: When I create a new User.ts file, it's really messing up the imports with User.d.ts. I'm correcting the StaffRole string in User.d.ts for now and we can continue using that one.

I also can't remove it from that file because it's being used in the Staff type in the same file

src/components/Admin/Edit.tsx
lspdrz: What exactly is the use of that? I can't pass that type up to the parent component. I think this is an ok case for any, because we do want to receive any type of values. The type constraints occur in the createStaff function in the parent component, EditStaff.

Also noting that antd uses any for these values in the above image

src/components/Admin/Edit.tsx
gabrielsaruhashi: I see what you're striving to acheive. In this case, the Edit staff page already exists. This is a placeholder solution for create staff
I think we're perceiving the scope of the task differently. These two pages are makeshift solutions, but they will eventually merge into our existing UI/UX that uses drawers as a way to create new entities.

src/components/Admin/EditStaff.tsx
gabrielsaruhashi: I believe we can if we type onSave using the generic <T> type that I proposed. That's the main use case for generic types

src/components/Admin/Edit.tsx
gabrielsaruhashi: antd's typing is not great
check comment below to understand the use case of generic types
Here's documentation that will help: https://www.typescriptlang.org/docs/handbook/2/generics.html

src/typings/Common.ts
```diff
@@ -29,3 +31,13 @@ export type Tab<TParam> = {
   key: TParam;
   tab: string;
 };
+
+export type FormItem = {
+  label: string;
+  name: string;
+  placeholder: any;
```
gabrielsaruhashi, line 38: should this be any?

src/typings/Common.ts
```diff
@@ -29,3 +31,13 @@ export type Tab<TParam> = {
...
+  required: boolean;
```
gabrielsaruhashi, line 33: might be convenient to make it optional
required?: boolean

src/typings/Common.ts
lspdrz: yes, it can be any type. String, boolean, array...

src/components/Admin/Edit.tsx
```diff
@@ -0,0 +1,79 @@
...
+          allowClear
+        >
+          {formItem.selectOptions.map((enumValue: string) => (
+            <Select.Option value={enumValue}>{enumValue}</Select.Option>
+          ))}
+        </Select>
+      );
+    }
+    switch (typeof formItem.placeholder) {
+      case "boolean":
```
gabrielsaruhashi, line 40: why not create a separate type field like the one we discussed? wouldn't it be more explicit to have a typeof and assigning a dummy placeholder name to identify a boolean?
This feels a bit hacky.

src/components/Admin/Edit.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const setValue = (value: any, name: string) =>
+    createForm.setFieldsValue({ [name]: value });
+
+  const getInputMessage = (formItem: FormItem) => {
+    if (formItem.selectOptions) return "Please select one of the options";
+    return `Please input your ${formItem.label.toLowerCase()}`;
+  };
+
+  const getInputBox = (formItem: FormItem) => {
+    if (formItem.selectOptions) {
```
gabrielsaruhashi, line 26: Same here:
Why not have a "type" in the FormItem for "select" like we discussed?
That'd be more clear typing + reading

src/redux/modules/staff.ts
gabrielsaruhashi: Im pretty positive u don't need this.
I grab error messages from payload.error.message

src/redux/modules/staff.ts
```diff
@@ -98,13 +106,25 @@ export const staffSlice = createSlice({
...
+        CREATE_STAFF,
+        `${action.payload.firstName} ${action.payload.lastName} was added successfully as a staff member of the facility! `,
+        "success"
+      );
+      staffAdapter.addOne(state, action.payload);
+    });
+    builder.addCase(createStaff.rejected, (state, action) => {
+      showToast(
+        CREATE_STAFF,
+        `Failed to create staff member: ${action.payload}`,
```
gabrielsaruhashi, line 120: have u tried action.error.message?

gabrielsaruhashi: when it throws u shouldn't use a payload

src/components/Admin/Edit.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+const Edit = ({ formItems, onCreate }: Props) => {
+  const [createForm] = Form.useForm();
+  const saveEntity = (values: EditFormValues) => {
+    onCreate(values);
+    createForm.resetFields(); // Would like to change this in the future to only reset upon successful POST
+  };
+
+  const setValue = (value: any, name: string) =>
```
gabrielsaruhashi, line 17: couldn't we type name to be 
`keyof FormItem`

src/typings/Common.ts
```diff
@@ -29,3 +31,13 @@ export type Tab<TParam> = {
...
+  selectOptions?: StaffRole[];
```
gabrielsaruhashi, line 34: We discussed this briefly on the previous PR, but I'd do something like

item: SelectFormItem | CheckboxFormItem | TextFieldFormItem

each of them can have a type and their specifc metadata e.g

SelectFormItem {
  type: 'select',
  options: string[]
} 

and have all the particular implementations in there instead of having multiple null fields / secret keywords

gabrielsaruhashi (changes requested):
I'm still wondering what we're gaining with this abstraction. I'd have rather started with a simple solution, and shipped that fast like we outlined on the ticket, instead of trying to abstract it right away and spending a week on it. 
These were supposed to be two temporary admin features. This Edit component won't be able to be reused for MeetingSlots, Kiosk creation etc.

src/components/Admin/Edit.tsx
lspdrz: This is really an implementation detail that would require a lot of changes and re-testing on my part. Worth doing instead of getting it into main for Andrew?

lspdrz:
> I'm still wondering what we're gaining with this abstraction. I'd have rather started with a simple solution, and shipped that fast like we outlined on the ticket, instead of trying to abstract it right away and spending a week on it. These were supposed to be two temporary admin features. This Edit component won't be able to be reused for MeetingSlots, Kiosk creation etc.

I went with the abstraction because I always abstract a component in React when I see/foresee it being reused, and it was supposed to be a simple abstraction. Was not supposed to take a week. But I agree this is not a productive workflow. It's going to take another hour at least to implement the next iteration of review comments and test them, which I will do tomorrow morning my time. I'm also going to try to write up a timeline of the work on this ticket so we can improve the workflow as this is not sustainable

src/components/Admin/Edit.tsx
lspdrz: Throwing type errors

lspdrz: image

src/typings/Common.ts
```diff
@@ -25,3 +25,28 @@ export type Tab<TParam> = {
...
 };
+
+export type TextFieldFormItem = {
+  type: 'string',
+  placeholder: string;
+}
+
+export type BooleanFormItem = {
+  type: 'boolean',
+  placeholder: boolean;
```
gabrielsaruhashi, line 36: what does placeholder mean for a boolean field?

src/components/Admin/Edit.tsx
```diff
@@ -0,0 +1,78 @@
...
+    if (formItem.item.type === 'enum') return "Please select one of the options";
+    return `Please input your ${formItem.label.toLowerCase()}`;
+  };
+
+  const getInputBox = (formItem: FormItem) => {
+    switch (formItem.item.type) {
+      case "boolean":
+        setValue(false, formItem.name); // Default checkbox values to false
+        return (
+          <Checkbox
```
gabrielsaruhashi, line 30: nit: 
<Checkbox  ... />

src/components/Admin/Edit.tsx
gabrielsaruhashi: I think that it would add too much debt.
The better approach imo would have been to go for a simple for for the two pages and then worked on a proper abstraction as fast-follows if/when we saw the need

src/components/Admin/EditStaff.tsx
```diff
@@ -0,0 +1,102 @@
...
+        selectOptions: ["admin", "staff", "investigator"],
+      },
+      required: true,
+    },
+    {
+      label: "Active?",
+      name: "isActive",
+      item: {
+        type: 'boolean',
+        placeholder: false,
```
gabrielsaruhashi, line 84: I think initialValue is a better name here

src/typings/Staff.ts
```diff
@@ -5,3 +5,28 @@ export interface Staff {
   facilityId: string;
```
gabrielsaruhashi, line 5: note that we need to update this type since the staff changes

gabrielsaruhashi (approved):

src/typings/Common.ts
lspdrz: it's not being used right now but it would allow us to set the default on the Checkbox

src/typings/Staff.ts
lspdrz: I don't understand what you mean here

PR #142, opened 2021-11-03
adds create facility page to doc admin

lspdrz:
[sc-1213]

DO NOT MERGE UNTIL https://github.com/Ameelio/connect-api-nest/pull/421 IS MERGED AND DEPLOYED

Adds the Create Facility Page to the DOC Admin

https://user-images.githubusercontent.com/9705113/140079185-3329c48a-541f-4ca2-bd7d-5dd332f561da.mov


src/typings/Common.ts
```diff
@@ -52,3 +55,20 @@ export type StaffData = {
   phone?: string;
   staffPosition: StaffPositionData;
 };
+
+export type FacilityData = {
```
gabrielsaruhashi, line 59: move to the page?
Seems like this is only used in there. 
We might want to use an Union type with Facility because a lot of these fields are covered in there

NewFacility = Facility & { publidId: string, ... }

src/redux/modules/staff.ts
```diff
@@ -54,6 +55,24 @@ export const createStaff = createAsyncThunk(
...
+  async (facilityData: FacilityData, { rejectWithValue }) => {
+    try {
+      const body = await fetchAuthenticated(`facility`, {
+        method: "POST",
+        body: JSON.stringify(facilityData),
+      });
+      const facility = body.data as Facility;
+
+      return facility;
+    } catch (error) {
```
gabrielsaruhashi, line 70: same comments as previous PRs

src/typings/Common.ts
lspdrz: Moved to Facility.ts like we did for staff creation

src/redux/modules/staff.ts
```diff
@@ -121,6 +142,21 @@ export const staffSlice = createSlice({
         error: action.error.message,
       };
     });
+    builder.addCase(createFacility.fulfilled, (state, action) => {
+      showToast(CREATE_FACILITY, "Facility successfully created!", "success");
+      // facilitiesAdapter.addOne(state, action.payload);
```
gabrielsaruhashi, line 147: ✂️

gabrielsaruhashi (approved):

PR #146, opened 2021-11-04
Login UX

cainlevy:
1. autoFocus the email field
2. leave the login form mounted in case credentials are wrong and need to be edited


src/pages/LoginPage/index.tsx
```diff
@@ -38,6 +38,8 @@ function UnconnectedLoginContainer({ session }: PropsFromRedux): ReactElement {
   }
 
   const onFinish = async (values: any) => {
+    if (session.status === "loading") return;
```
cainlevy, line 41: apparently we're on our own to prevent double submits 🤯 

(you can bypass the button by pressing enter)

gabrielsaruhashi: heh guess react hook form is coming sooner than I expected

gabrielsaruhashi (approved):

PR #149, opened 2021-11-05
more strict typing patterns for route navigation

gabrielsaruhashi:
The previous navigation pattern was very bug-prone. It accepted any string.


package.json
```diff
@@ -96,14 +96,21 @@
...
-    "eslint-config-airbnb": "^18.2.0",
+    "eslint-config-airbnb": "^18.2.1",
     "gh-pages": "^3.1.0",
     "husky": "^4.2.5",
     "lint-staged": "^10.2.11",
     "postcss": "^7.0.35",
     "prettier": "^2.0.5",
     "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.1.2",
-    "typescript": "^3.9.7"
+    "typescript": "^4.4.4"
```
gabrielsaruhashi, line 109: template literals were introduced on TS 4.1

package.json
```diff
@@ -96,14 +96,21 @@
...
+  },
+  "resolutions": {
+    "**/typescript": "^4.4.4",
+    "**/@typescript-eslint/eslint-plugin": "^4.15.1",
+    "**/@typescript-eslint/parser": "^4.15.1"
```
cainlevy, line 114: 1) wait this works with npm??
2) why not just specify these versions in the dependencies?

src/constants/Routes.ts
```diff
@@ -50,5 +76,10 @@ export const ROUTES: Route[] = [
   { path: "/admin", component: AdminPage, label: "Admin" },
   { path: "/call/:id", component: RecordingPage, label: "Past Call" },
   { path: "/visitors", component: VisitorsPage, label: "Visitors" },
+  {
+    path: "/message-review",
+    component: MessageReviewPage,
+    label: "Review Messages",
+  },
```
cainlevy, line 83: feature flagged?

cainlevy (approved):
cool use of template literal types

package.json
gabrielsaruhashi: I was struggling to get template literals to compile 😱 I thought this had resolved it but updating typescript to the right version was the right answer. 
Thanks for calling this out

PR #151, opened 2021-11-09
Apollo client setup

gabrielsaruhashi:
👋 redux


src/api/client.ts
```diff
@@ -1,13 +1,22 @@
-import { createClient, fetchExchange } from "urql";
```
cainlevy, line 1: can we drop `urql`?

cainlevy (approved):
👋 

long goodbye

src/api/client.ts
gabrielsaruhashi: yep! will drop it on the next PR

PR #152, opened 2021-11-09
pending messages review page: loading connections & fetching messages

gabrielsaruhashi:
## Proposed changes
Foundational work for the pending messages review page.
image


src/graphql/GetConnection.graphql
```diff
@@ -0,0 +1,11 @@
+query GetConnection($id: ID!) {
```
cainlevy, line 1: GetConnectionMessages? GetConnectionTimeline?

src/pages/MessageReviewPage.tsx
```diff
@@ -1,15 +1,90 @@
...
 
 interface Props {}
 
 const MessageReviewPage = (props: Props) => {
+  const facilityId = useAppSelector((state) => state.facilities.selected?.id);
+
+  const { error: pendingMessageError, data: pendingMessageData } = useQuery(
+    GetPendingMessagesDocument,
+    {
+      variables: { id: facilityId || "" },
```
cainlevy, line 23: we should eliminate the null case of `facilityId` some other way. maybe throw an error? this would cause an error anyway, but a less helpful one.

src/graphql/GetConnection.graphql
```diff
@@ -0,0 +1,11 @@
...
+  connection(id: $id) {
+    messages {
+      edges {
+        node {
+          ...MessageFragment
```
cainlevy, line 6: i have a feeling that we could get into trouble with fragments that load nested relations. it makes them less reusable and we'll have to watch out for circular loading.

if we use small fragments we can always compose them:

```suggestion
          ...MessageFragment
          connection {
            ...ConnectionFragment
            inmate {
              ...InmateFragment
            }
            visitor {
              ...VisitorFragment
            }
          }
```

src/pages/MessageReviewPage.tsx
```diff
@@ -1,15 +1,90 @@
...
+    );
+    const connections: TConnectionSidebarItem[] = uniqBy(
+      messages.map((m) => m.connection),
+      (c) => c.id
+    ).map((c) => ({
+      id: c.id,
+      inmateName: genFullName(c.inmate),
+      visitorName: genFullName(c.visitor),
+    }));
+    setConnections(connections);
```
cainlevy, line 49: i realize this is just a way of getting started but we should build this timeline in the backend, even if it's composited together from multiple database queries for now. our goal should be a timeline that can paginate.

cainlevy (changes requested):

src/pages/MessageReviewPage.tsx
gabrielsaruhashi: i see makes sense! 
Do you agree with the typing strategy? @cainlevy 
I'll just move the types the backend and do the conversions backend side
I am unfortunately tied up all day today, but I will have time tmr.

cainlevy: yeah the union type makes sense to me! we'll end up publishing it over graphql and then referencing `__typename` to narrow the types during render.

cainlevy: https://www.apollographql.com/docs/apollo-server/schema/unions-interfaces/

gabrielsaruhashi: awesome. on it

src/components/MessageReviewPanel/ConnectionListSidebar/ConnectionListSidebar.tsx
```diff
@@ -0,0 +1,41 @@
+import { UserOutlined } from "@ant-design/icons";
+import { Layout, Typography } from "antd";
+import React from "react";
+import { Inmate, Visitor } from "src/api/graphql";
+import { genFullName } from "src/utils";
+import ConnectionListItem from "./ConnectionListItem";
+
+export type TConnectionListItem = {
```
cainlevy, line 8: Why "T"? Is that a local convention?

src/graphql/fragments/MessageFragment.graphql
```diff
@@ -0,0 +1,12 @@
+fragment MessageFragment on Message {
+  id
+  detectedKeywords {
+    id
+    spelling
+  }
+  status
+  createdAt
+  availableAt
+  senderId
```
cainlevy, line 10: What value is this? Is it a malan ID? Are we exposing those to the API?

The answer may be in the backend but ... phone.

src/graphql/fragments/VisitorFragment.graphql
```diff
@@ -0,0 +1,6 @@
+fragment VisitorFragment on Visitor {
+  id
+  firstName
+  lastName
+  malanId
```
cainlevy, line 5: What's the use case for malanId in the frontend? I'm hopeful we can hide it behind our GraphQL API. Maybe trying to put the cat back in the bag here.

src/pages/MessageReviewPage.tsx
```diff
@@ -1,15 +1,77 @@
...
+  const [
+    queryConnection,
+    { error: queryConnectionError, data: queryConnectionData },
+  ] = useLazyQuery(GetConnectionMessagesDocument);
+
+  const [selected, setSelected] = useState<TConnectionListItem>();
+  const [connections, setConnections] = useState<TConnectionListItem[]>([]);
+  const [messages, setMessages] = useState<TMessageItem[]>([]);
+
+  useEffect(() => {
```
cainlevy, line 33: This looks like a useMemo.

But really it's going to save very little. This component should re render very rarely as staff select new connections.

src/pages/MessageReviewPage.tsx
```diff
@@ -1,15 +1,77 @@
...
+      ({ node }) => node
+    );
+    const uniqConnections = uniqBy(
+      messages.map((m) => m.connection),
+      (c) => c.id
+    );
+    setConnections(uniqConnections);
+  }, [pendingMessageData]);
+
+  useEffect(() => {
```
cainlevy, line 45: Looks like a useMemo (or just bare JS, really)

src/pages/MessageReviewPage.tsx
```diff
@@ -1,15 +1,77 @@
...
+
+    setMessages(messages);
+  }, [queryConnectionData]);
+
+  if (pendingMessageError) throw new Error(pendingMessageError.message);
+  if (queryConnectionError) throw new Error(queryConnectionError.message);
+
+  const onSelect = (connection: TConnectionListItem) => {
+    setSelected(connection);
+    queryConnection({ variables: { id: connection.id } });
```
cainlevy, line 59: I suspect you could also have a standard useQuery and let Apollo notice when the argument changes. There's even a "skip" option for when the current value is null.

I don't know if one implementation is better. We have options!

cainlevy (approved):

src/components/MessageReviewPanel/ConnectionListSidebar/ConnectionListSidebar.tsx
gabrielsaruhashi: I personally like doing this when I have a component-level props that is exportable / reusable elsewhere
It make more clear the relationship between the two

src/graphql/fragments/MessageFragment.graphql
gabrielsaruhashi: Yep! Malan ID
The service function for createMessage that I created ensures we use malan

src/graphql/fragments/VisitorFragment.graphql
gabrielsaruhashi: We'd use to detect whether the sender is a user or an inmate
We can do that conversion at the resolver level, though

src/pages/MessageReviewPage.tsx
gabrielsaruhashi: oh interesting! 
I liked this one because it makes it pretty clear that the query is reactive to a UI interaction
happy to explore more on fast-follow PRs

src/pages/MessageReviewPage.tsx
gabrielsaruhashi: idk im always so wary of memoization for inexpensive computations lol, but I agree it's more clear. Cool with just doing that for useStart / useEffect combos

PR #154, opened 2021-11-12
Keyword search field and filtering

wittejm:
looks like this:
search_field_screenie


cainlevy (approved):
that looks like it should work! 👍

PR #155, opened 2021-11-15
Jordanwitte/sc 1486/keyword settings table is expandable displays

wittejm:
First commit is just to update `graphql.ts` before generating my own required graphql changes. 

Screenies showing the horizontal scroll:
Screenshot 2021-11-15 at 14-03-22 Ameelio Connect Video Conferencing
Screenshot 2021-11-15 at 14-05-40 Ameelio Connect Video Conferencing

And the tooltip: 
global_keyword_tooltip


src/graphql/GetMessageKeywords.graphql
```diff
@@ -7,6 +7,20 @@ query GetMessageKeywords($id: ID!) {
         description
         spellings
         isGlobal
+        createdAt
+        createdBy {
+          id
+          firstName
+          lastName
+          malanId
```
wittejm, line 15: My instinct was to exclude malanId from this query but it is apparently required on the Staff type generated in graphql.ts and generates a type error without.

src/components/Settings/KeywordManagementSettings.tsx
```diff
@@ -18,6 +19,11 @@ const KeywordManagementSettings = (props: Props) => {
   const computeMatch = (searchQuery: string, keyword: MessageKeyword) => {
     return isSubstring(searchQuery, keyword.name) || keyword.spellings.map(s => isSubstring(searchQuery, s)).some(e => e);
   }
+  const globalTooltip = (() => {
+    return <div className="flex" >Global <Tooltip title="The global keywords list is maintained by Ameelio.">
+      <InfoCircleOutlined className="p-1" style={{ color: "gray" }} />
+    </Tooltip></div>
+  })
```
cainlevy, line 26: may as well hoist it all the way out as a static react component

src/components/Settings/KeywordManagementSettings.tsx
```diff
@@ -31,33 +37,68 @@ const KeywordManagementSettings = (props: Props) => {
...
       />
       <Table
-        dataSource={filteredKeywords}
+        dataSource={filteredKeywords.map((keyword, index) => ({
+          key: index,
+          ...keyword,
+          createdBy: !!keyword.createdBy ? `${keyword.createdBy.firstName} ${keyword.createdBy.lastName}` : "",
+          createdAt: !!keyword.createdAt ? formatDate(keyword.createdAt, "date") : "-",
+          updatedBy: !!keyword.lastUpdatedBy ? `${keyword.lastUpdatedBy.firstName} ${keyword.lastUpdatedBy.lastName}` : "-",
+          updatedAt: !!keyword.updatedAt ? formatDate(keyword.updatedAt, "date") : "-",
```
cainlevy, line 53: did you also consider writing this logic in the column renders? what's the tradeoff?

src/graphql/GetMessageKeywords.graphql
cainlevy: > required on the Staff type generated in graphql.ts

I'm not following. The `graphql.ts` file might say that it's non-nullable, but that shouldn't require you to fetch the field. Maybe there's a component where you're using the data that requires `malanId`?

cainlevy (approved):

src/components/Settings/KeywordManagementSettings.tsx
wittejm: Oh nice. On first pass I was thinking that like table cells wanted string values, but clearly they can take objects and indeed I can push this into the column render.

PR #165, opened 2021-11-18
Use absolute (not paginated) index to load correct entry in modal.

wittejm:
The built-in `index` value when rendering a table is the index on the current page, so it's always 0-9. But we want the absolute index for the entire dataset. This was a bug that clicking a keyword on the second or later page always loaded a keyword from the first page instead.


src/components/Settings/KeywordManagementSettings.tsx
```diff
@@ -109,11 +110,11 @@ const KeywordManagementSettings = (props: Props) => {
...
           title="Operations"
           key="action"
-          render={(_, __, index) => (
-            isUserAdmin && !data?.facility.keywords[index].isGlobal ?
+          render={(value) => (
+            isUserAdmin && !data?.facility.keywords[value.index].isGlobal ?
               <div className="flex gap-2">
                 <Typography.Link onClick={() => {
-                  setEditingKeywordIndex(index)
+                  setEditingKeywordIndex(value.index);
```
gabrielsaruhashi, line 117: i didnt follow the previous reviews but why store the index of the editing keyword instead of the keyword itself?

wittejm: That's a good point. I think I can take it out and it will be cleaner.

src/components/Settings/KeywordManagementSettings.tsx
```diff
@@ -128,11 +127,11 @@ const KeywordManagementSettings = (props: Props) => {
...
         />
       }
-      {editingKeywordIndex > -1 &&
+      {!!editingKeyword &&
         <AddEditKeywordModal
-          keyword={data?.facility.keywords[editingKeywordIndex]}
-          existingKeywordNames={data?.facility.keywords.map(k => k.name).filter((n => n != data.facility.keywords[editingKeywordIndex].name)) || []}
-          onClose={() => setEditingKeywordIndex(-1)}
+          keyword={editingKeyword}
+          existingKeywordNames={data?.facility.keywords.map(k => k.name).filter((n => n != editingKeyword.name)) || []}
```
gabrielsaruhashi, line 133: i went through this file and it seems like u do this data?.facility.keywords in a couple of places

i'd just declare it as a const.

const keywords = data?.facility.keywords || []

instead of doing the check multiple times in different places

gabrielsaruhashi (approved):

src/components/Settings/KeywordManagementSettings.tsx
wittejm: Makes sense. But `data` loads asynchronously from the graphql result, so computing the keyword list at the top of the render results in an empty keyword list. 

I tried following your suggestion but put it in a separate PR here, because I think it required adding a new state variable and a second useEffect call, which seems a little icky:  https://github.com/Ameelio/connect-doc-client/pull/167

gabrielsaruhashi: sounds good! feel free to merge :) it's a nit

PR #166, opened 2021-11-18
Revert "Revert "Connection Drawer""

lspdrz:
Reverts Ameelio/connect-doc-client#162


src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -51,10 +60,50 @@ const ConnectItemButton = ({
...
+      'M/d/y H:mm'
+    )}`,
+    `${format(
+      new Date(restriction.expiresAt),
+      'M/d/y H:mm'
+    )}`,
+    `${restriction.reason}`
+  ]
+  return contents.map(content => (
+    <Descriptions.Item label='Disabled by' labelStyle={{
```
gabrielsaruhashi, line 77: i believe react will complain about a missing key

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -51,10 +60,50 @@ const ConnectItemButton = ({
   }
 };
 
+const getRestrictionDetails = (restriction: ConnectionRestriction) => {
+  const contents = [
```
gabrielsaruhashi, line 64: Why not just treat it as a single React.ReactNode instead of an array of React.ReactNode with a single item -- i.e I think u can just move this inside the description.item

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -67,38 +116,92 @@ const ConnectionItem: React.FC<Props> = ({ connection, type, onUpdate }) => {
...
           </Descriptions>
-          <ConnectItemButton
-            className="ml-auto"
-            onUpdate={onUpdate}
-            status={connection.status}
-          />
+          {featureEnabled("EMESSAGING") && connectionRestrictions && (
+            <Descriptions column={1}>
+              {/*
+              TODO: Re-introduce when we have an emessaging meetingtype
```
gabrielsaruhashi, line 164: link to shortcut story?

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -67,38 +116,92 @@ const ConnectionItem: React.FC<Props> = ({ connection, type, onUpdate }) => {
...
+        title={drawerTitle}
+        width={480}
+        onClose={() => setOpen(false)}
       >
         <div className="flex flex-col">
-          <Descriptions title="Connection Info" column={1}>
-            <Descriptions.Item label="Connection ID">
-              {connection.id}
+          <Descriptions column={1}>
+            <Descriptions.Item label="Incarcerated Person" labelStyle={labelStyle}>
```
gabrielsaruhashi, line 127: ideally we would override the antd theme, but this works for now

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -67,38 +116,92 @@ const ConnectionItem: React.FC<Props> = ({ connection, type, onUpdate }) => {
...
+          <Descriptions column={1}>
+            <Descriptions.Item label="Contact" labelStyle={labelStyle}>
               <ContactCell contact={connection.contact} />
             </Descriptions.Item>
-            <Descriptions.Item label="Created At">
+            <Descriptions.Item label="Relationship to Incarcerated" labelStyle={labelStyle}>
+              {`${capitalize(connection.relationship)}`}
+            </Descriptions.Item>
+            <Descriptions.Item label="Email" labelStyle={labelStyle}>
+              {`${connection.contact.email}`}
```
gabrielsaruhashi, line 143: instead of applying label style N times for each of the n rows, I think we should move this to a wrapper <DescriptionItem  /> that does this for us

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -67,38 +116,92 @@ const ConnectionItem: React.FC<Props> = ({ connection, type, onUpdate }) => {
...
+              {inPersonContactRestriction && getRestrictionDetails(inPersonContactRestriction)}
+              <Descriptions.Item label='Allowed in-person visits, without contact' labelStyle={labelStyle}>
+                {inPersonNoContactRestriction ? 'No' : 'Yes'}
+              </Descriptions.Item>
+              {inPersonNoContactRestriction && getRestrictionDetails(inPersonNoContactRestriction)}
+            </Descriptions>
+          )}
+          <Divider/>
+          <Space className='flex justify-end'>
+            <Row className=''>
```
gabrielsaruhashi, line 185: why wrap the buttons with Rows? You can probably just have the buttons here

src/components/Requests/StaffCell.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const StaffCell = ({ staff }: Props) => {
+  const navigate = useAppNavigate();
+  if (!staff) return null
+  return (
+    <div className="flex gap-x-4">
+      {/* TODO: reintroduce when we have individual staff pages
+      <Typography.Link onClick={() => navigate(`/staff/${staff.id}`)}> */}
+      <Typography.Link onClick={() => navigate(`/staff`)}>
+        {`${staff.firstName} ${staff.lastName}`}
```
gabrielsaruhashi, line 19: genFullName(staff)

gabrielsaruhashi (changes requested):
left a couple of comments as i reviewed

src/components/Profile/Connections/ConnectionItem.tsx
lspdrz: Do you mean create four description.items and return them? that's what I had before and Jordan rightfully pointed out it wasn't very DRY, so now we're mapping the array

src/components/Profile/Connections/ConnectionItem.tsx
lspdrz: Will sort out when we solve above

src/components/Profile/Connections/ConnectionItem.tsx
gabrielsaruhashi: I see I think this array with a JSX.element and two strings is what's confusing. Could you just move them inside the return JSX?

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -51,10 +60,50 @@ const ConnectItemButton = ({
   }
 };
 
+const getRestrictionDetails = (restriction: ConnectionRestriction) => {
```
gabrielsaruhashi, line 63: this is returning JSX so I'd just follow standard component naming
<RestrictionDetails />

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -67,38 +116,92 @@ const ConnectionItem: React.FC<Props> = ({ connection, type, onUpdate }) => {
...
+              {/*
+              TODO: Re-introduce when we have an emessaging meetingtype
+              <Descriptions.Item label='Allowed video calls' labelStyle={labelStyle}>
+                {eMessagingRestriction ? 'No' : 'yes'}
+              </Descriptions.Item>
+              {callRestriction && getRestrictionDetails(eMessagingRestriction)}*/}
+              <Descriptions.Item label='Allowed video calls' labelStyle={labelStyle}>
+                {callRestriction ? 'No' : 'Yes'}
+              </Descriptions.Item>
+              {callRestriction && getRestrictionDetails(callRestriction)}
```
gabrielsaruhashi, line 172: then here you do <RestrictionDetails/> instead of doing this function notation

src/components/Profile/Connections/ConnectionItem.tsx
lspdrz: CAn't do this because wrapper components don't work for Description.Item: https://github.com/ant-design/ant-design/issues/27411

src/components/Profile/Connections/ConnectionItem.tsx
lspdrz: CAn't do this because wrapper components don't work for Description.Item: ant-design/ant-design#27411

src/components/Profile/Connections/ConnectionItem.tsx
lspdrz: CAn't do this because wrapper components don't work for Description.Item: ant-design/ant-design#27411

src/components/Profile/Connections/ConnectionItem.tsx
lspdrz: CAn't do this because wrapper components don't work for Description.Item: ant-design/ant-design#27411

gabrielsaruhashi (approved):

PR #169, opened 2021-11-18
highlights detected keywords

gabrielsaruhashi:
image


src/utils/UI.ts
```diff
@@ -35,3 +35,8 @@ export const showToast = (
       break;
   }
 };
+
+export const makeBold = (item: string, keywords: string[]) => {
+  var re = new RegExp(keywords.join('|'), "g");
```
cainlevy, line 40: @wittejm any suggestions for how we can highlight as carefully as we tokenize? what's a reasonable effort?

src/components/MessageReviewPanel/MessageCard.tsx
```diff
@@ -54,11 +55,26 @@ const MessageCard: React.FC<Props> = ({ message, className }: Props) => {
         }`}
       >
         <MessageStatusIcon />
-        <div className={`border border-gray rounded bg-white p-2`}>
-          <Typography.Text>{message.content}</Typography.Text>
+        <div className={`border border-gray rounded bg-white grid gap-2 p-2`}>
+          <Typography.Text>
+            {message.detectedKeywords.length > 0
+              ? ReactHtmlParser(
```
cainlevy, line 61: looks like a good library for the situation!

cainlevy (approved):

src/utils/UI.ts
gabrielsaruhashi: @wittejm merging it in, but curious to hear your follow-up
My expectation here is that our de-normalized approach would allow us to store the exact match in the "spelling" column of the MessageDetectedKeyword. An exact match would theoretically make this simple regex pattern match expression work

wittejm: We do keep the exact string matches, but I think your regex as written will erroneously highlight substrings. Also we want to be able to refer to the underlying keyword for each highlighted word. 

Since tokens can be either split on non-alphanumeric characters, or include them, pattern matching for them could be tough.

New idea: what if in MessageDetectedKeyword we save the location of the keyword as the character count of the beginning of the word?

cainlevy: What if a message contains multiple of the same keyword? Would we create multiple MessageDetectedKeyword, or store multiple locations?

wittejm: Good question. Probably multiple MessageDetectedKeyword.

PR #170, opened 2021-11-19
delete keyword with confirmation modal

wittejm:
delete_confirm_modal

Extracted the edit confirmation popover content to reuse here.

This currently has the issue that double-clicking the delete-confirmation will throw an error because the second delete attempt fails. I'd like to catch the error in the backend and return success=false in the response, but when I tried wrapping the delete service function in try/catch, that seemed to have no effect and the error is still thrown. I don't see any other examples where our graphql mutations catch errors to return nicer failures. I also tried wrapping the mutation call in the frontend in try/catch and that had no effect either.  @cainlevy maybe you could take a look at this with me?


src/components/Settings/KeywordManagementSettings.tsx
```diff
@@ -26,6 +28,8 @@ const KeywordManagementSettings = (props: Props) => {
   const [filteredKeywords, setFilteredKeywords] = useState<MessageKeyword[]>(data ? data.facility.keywords : []);
   const [isAddingNewKeyword, setIsAddingNewKeyword] = useState(false);
   const [editingKeyword, setEditingKeyword] = useState<MessageKeyword>();
+  const [deletingKeyword, setDeletingKeyword] = useState<MessageKeyword>();
+  const [deleteKeywordMutation] = useMutation(DeleteLocalMessageKeywordDocument);
```
cainlevy, line 32: i just realized that i've seen @gabrielsaruhashi send actions like this all the way up to the page component. @gabrielsaruhashi is that our pattern? does it also apply to data fetching with `useQuery`?

cainlevy (approved):

PR #173, opened 2021-11-23
Disable submit/confirm buttons on click

wittejm:
To prevent double-click errors.


src/components/Settings/AddEditKeywordModal.tsx
```diff
@@ -26,6 +26,7 @@ const AddEditKeywordModal = ({
   const [spellings, setSpellings] = useState(keyword?.spellings?.join(", ") || "");
   const [category, setCategory] = useState(keyword?.category);
   const [confirmEditVisible, setConfirmEditVisible] = useState(false);
+  const [saveButtonDisabled, setSaveButtonDisabled] = useState(false);
```
gabrielsaruhashi, line 29: im pretty sure useMutation returns this information.
Have you checked the documentation?

gabrielsaruhashi: https://www.apollographql.com/docs/react/data/mutations/#executing-a-mutation

gabrielsaruhashi: You don't need to track this independently

src/components/Settings/ConfirmOverride.tsx
```diff
@@ -15,18 +15,26 @@ const ConfirmOverride = ({
...
+}: Props) => {
+  const [confirmButtonDisabled, setConfirmButtonDisabled] = useState(false);
+  return (
+    <div className="flex" style={{ maxWidth }}>
+      <InfoCircleFilled className="pt-3 pr-3" style={{ color: "#FAAD14", verticalAlign: 'middle' }} />
+      <div>
+        You're about to override an action taken by {previousEditorFullName}. Would you like to continue?
+        <Divider />
+        <div className="flex gap-2">
+          <Button style={{ "marginLeft": "auto" }} onClick={onClickNo}>No</Button>
```
gabrielsaruhashi, line 27: We prefer using tailwind whenever possible. 
className="ml-auto"

src/components/Settings/ConfirmOverride.tsx
```diff
@@ -15,18 +15,26 @@ const ConfirmOverride = ({
...
-      <Divider />
-      <div className="flex gap-2">
-        <Button style={{ "marginLeft": "auto" }} onClick={onClickNo}>No</Button>
-        <Button type="primary" htmlType="submit"
-          onClick={onClickYes} >Yes</Button>
+}: Props) => {
+  const [confirmButtonDisabled, setConfirmButtonDisabled] = useState(false);
+  return (
+    <div className="flex" style={{ maxWidth }}>
+      <InfoCircleFilled className="pt-3 pr-3" style={{ color: "#FAAD14", verticalAlign: 'middle' }} />
```
gabrielsaruhashi, line 22: tailwind? 
Tailwind has a 100-900 color system as well

gabrielsaruhashi (changes requested):

cainlevy:
Some day I hope we invest in components that make it easier to remember things like this. For example, we could create some Button wrapper that requires us to pass a `loading: boolean` prop.

src/components/Settings/ConfirmOverride.tsx
wittejm: The color here "Calendula Gold / 6" doesn't have a match in Tailwind. Approximate with ~Amber 600  or stick with the design?

src/components/Settings/ConfirmOverride.tsx
```diff
@@ -8,25 +8,33 @@ interface Props {
...
-      You're about to override an action taken by {previousEditorFullName}. Would you like to continue?
-      <Divider />
-      <div className="flex gap-2">
-        <Button style={{ "marginLeft": "auto" }} onClick={onClickNo}>No</Button>
-        <Button type="primary" htmlType="submit"
-          onClick={onClickYes} >Yes</Button>
+  disabled
+}: Props) => {
+  return (
+    <div className="flex" style={{ maxWidth }}>
```
wittejm, line 22: I think I want to stick with `style={{}}` here because the tailwind smallest maxWidth is too big.

gabrielsaruhashi: makes sense -

gabrielsaruhashi: curious about why we need the max width - is that to make sure we wrap the text in case the previousEditorFullName is long?

gabrielsaruhashi: If ye it might make sense to just pick a fixed size in rem?
e.g "w-48"

gabrielsaruhashi (approved):

src/components/Settings/ConfirmOverride.tsx
wittejm: It's just really small as a popover to confirm an edit and larger as a modal to confirm a delete. There's actually no reason to use maxWidth over width, and that's cool, I didn't realize width rem could be anything using tailwind.

PR #176, opened 2021-11-24
DOC emessaging features only present if hasMessaging=true

wittejm:
The change I made in commit 3 saves a single line and looks kind of cute but I'm not sure if it's an abuse of the `!` operator.


src/components/Menu/Menu.tsx
```diff
@@ -132,7 +132,7 @@ export default function Menu({
             Meeting Requests
             <Badge count={callRequestsCount} />
           </AntdMenu.Item>
-          {featureEnabled("EMESSAGING") && (
+          {featureEnabled("EMESSAGING") && selected.hasMessaging && (
```
gabrielsaruhashi, line 135: This works for now!

Should we just move it to a util perhaps that takes the facility as a param in case we ever change the implementation / representation of messaging?

I think in the long run we might want to consider using a react hook that returns all the services that are available for that facility

src/pages/MessageReviewPage.tsx
```diff
@@ -141,48 +142,50 @@ const MessageReviewPage = (props: Props) => {
     [];
 
   return (
```
gabrielsaruhashi, line 144: Instead of doing it here, we probably want to do this at the Route-level. I had done something similar with a feature flag, @cainlevy mentioned doing at the router level is generally enough

That'd require some refactoring potentially given how we declare routes right now, so im fine with this as a stepping stone

src/components/Menu/Menu.tsx
wittejm: I think a util function might look like `facilityHasFeature(facility, feature)` but we don't currently have support for any other customizable features-- it's something talked about only generally so far AFAIK. When we do build that generalization of features probably including an enum of feature names, then I think that's a good time to write that util function and refactor these flags, rather than try to future-proof these flags now. 

Or maybe that's what you meant by "This works for now" and everything you said is a suggestion for later?  :slightly_smiling_face:

src/pages/MessageReviewPage.tsx
```diff
@@ -141,48 +142,50 @@ const MessageReviewPage = (props: Props) => {
     [];
```
gabrielsaruhashi, line 143: could we please just do 
if (!facilityhasMessaging) return <Redirect to={{ pathname: "/" }} />

I personally feel this is more readable that the long ternary operator

gabrielsaruhashi (approved):

src/components/Menu/Menu.tsx
cainlevy: I was going to suggest that we remove `featureEnabled(...)` and rely on the entity flag.

The future of feature flags is server controlled, anyway. This is just a permanent version.

cainlevy (commented):
Note that we have to enforce this bit in the backend as well, in all relevant mutations. Would you make sure that has a ticket?

wittejm:
Getting rid of the featureEnabled flag meant making the message-review route always present and relying entirely on the conditional <Redirect> element. Does that seem ok for now?

PR #177, opened 2021-11-25
replaced feature flag with hasMessaging boolean

lspdrz:


gabrielsaruhashi:
@lspdrz merging this in because it's throwing errors in main

gabrielsaruhashi (approved):

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -92,6 +91,8 @@ const ConnectionItem: React.FC<Props> = ({ connection, type, onUpdate }) => {
   const { loading, error, data } = useQuery(GetConnectionConnectionRestrictionsDocument, {
     variables: { id: connection.id || "" },
   });
+  const { hasMessaging: facilityhasMessaging } = useAppSelector((state) => state.facilities.selected!);
```
gabrielsaruhashi, line 94: im a little torn because only higher-level containers and HOCs should have access to the redux store. This is technically an anti-pattern in redux

gabrielsaruhashi (changes requested):
actually, @lspdrz I'll refactor this. 
It's generally considered bad practice to connect child components to store.
I'll create a custom hook for this

PR #178, opened 2021-11-26
react hook for facility features

gabrielsaruhashi:
## Description
It's generally considered an anti-pattern in redux to have child components connect to the global state.

Only higher-level containers and HOCs should have such access.

Having a dedicated hooks that decides whether or not a facility has a given feature seems like the more scalable way to build this out. This can contain more complicated logic in case we move away from a simple boolean feature flag in the entity model.


lspdrz (approved):

cainlevy (approved):
> It's generally considered an anti-pattern in redux to have child components connect to the global state.

does this mean:

a) files in src/components should not select from redux
b) files in src/components should not select from global UI state (like the currently selected facility)?

PR #180, opened 2021-11-30
adjusted copy for cancellation reasons

gabrielsaruhashi:


gabrielsaruhashi:
merging this one in as it's a simple copy change :)

PR #183, opened 2021-12-01
connection restriction modal

lspdrz:
Do not merge until https://github.com/Ameelio/connect-api-nest/pull/541 is merged and deployed


src/graphql/DeleteConnectionRestriction.graphql
```diff
@@ -0,0 +1,7 @@
+mutation DeleteConnectionRestriction(
+  $input: DeleteConnectionRestrictionInput!
+) {
+  deleteConnectionRestriction(input: $input) {
+    success
```
gabrielsaruhashi, line 5: You grab the ID of the connection restriction here so that you can update the cache.
@cainlevy as this is a delete, I imagine he doesn't need the complete record right. ID suffices to avoid a 2nd query

src/graphql/CreateConnectionRestriction.graphql
```diff
@@ -0,0 +1,17 @@
+mutation CreateConnectionRestriction(
+  $input: CreateConnectionRestrictionInput!
+) {
+  createConnectionRestriction(input: $input) {
+    connectionRestriction {
```
gabrielsaruhashi, line 5: seems like connection restriction is used in a couple of places
let's create a fragment to avoid duplication?

src/graphql/UpdateConnectionRestriction.graphql
```diff
@@ -0,0 +1,12 @@
+mutation UpdateConnectionRestriction(
+  $input: UpdateConnectionRestrictionInput!
+) {
+  updateConnectionRestriction(input: $input) {
+    connectionRestriction {
+      id
```
gabrielsaruhashi, line 6: would use the fragment here as well

src/components/Modals/EditPermissionsModal/EditPermissionsModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+                                connectionRestrictionId: restriction.id,
+                                expiresAt,
+                                reason
+                            }
+                        }
+                    }
+                )
+            }
+        } else {
+            if (isEditing) {
```
wittejm, line 105: When does this conditional evaluate to false?

lspdrz: Whenever a DOC user opens the modal but doesn't actually start adding a new permission. And yeah, I think it will

src/graphql/DeleteConnectionRestriction.graphql
lspdrz: That would require a backend change, which I'm happy to do, but sort of goes against the pattern Jordan and I sort of silently established with the deletes.

For this particular case, I'll already have the connectionRestriction id (it's in the restriction prop passed down) so I don't think I'll have to query for it from graphql

lspdrz: Similar to how Jordan does on line 31 of KeywordManagementSettings.tsx

src/components/Modals/EditPermissionsModal/EditPermissionsModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+                    reason: restriction?.reason || ""
+                }}
+            >
+                <Row className="flex justify-between">
+                    <Row>
+                        <InmateCell inmate={inmate}/>
+                        {` & `}
+                        <ContactCell contact={contact}/>
+                    </Row>
+                    <Switch defaultChecked={!isEditing} onChange={() => setIsEditing(!isEditing)} />
```
cainlevy, line 147: i'm finding it challenging to understand the `isEditing` boolean and all of the ways this modal gets used.

- `isEditing` starts with `!!restriction` and changes with this toggle
- if you open the dialog on an unrestricted permission, you are not editing. this switch is on.
- if you open the dialog on a restricted permission, you are editing. the switch is off. if you switch it off, then submitting will trigger a delete.

okay the behavior makes sense when i type it out, it just takes some effort. i wonder if it's a naming thing? 🤔

cainlevy (approved):
this is a trickier modal than i've seen in a while, with the three blended edit modes

wittejm (approved):

lspdrz:
> this is a trickier modal than i've seen in a while, with the three blended edit modes

Haha yeah but the logic's a little fun

PR #184, opened 2021-12-05
calls bug fix

lspdrz:
DO NOT MERGE UNTIL https://github.com/Ameelio/connect-doc-client/pull/184 is merged and deployed

[sc-1329]

Fixes the calls and visits tables on the Visitor and Inmate pages to include _all_ calls for that visitor/inmate. Previously only calls within the current month were showing.

image
Screen Shot 2021-12-06 at 23 56 04


src/utils/Message.ts
```diff
@@ -17,7 +17,7 @@ export function getMessageStatusLabel(status: MessageStatus) {
 }
 
 export function getMessageSenderType(message: TMessageItem) {
-  return message.senderId === message.connection.inmate.malanId
+  return message.sender.id === message.connection.inmate.malanId
```
wittejm, line 20: We can get the type name directly now, like so:
https://github.com/Ameelio/pathways-client/blob/37d692bef1eeea88a54d2f60fd39d7813367cd30/src/utils/Message.ts#L20

wittejm (approved):

src/graphql/GetInmate.graphql
```diff
@@ -0,0 +1,26 @@
+query GetInmate($inmateId: ID!) {
+  inmate(id: $inmateId) {
+    ...InmateFragment
+    connections {
+      inmate {
+        ...InmateFragment
+      }
```
cainlevy, line 7: do you mean `visitor` here? we already have the inmate by id.

src/graphql/fragments/MeetingFragment.graphql
```diff
@@ -0,0 +1,59 @@
...
+    }
+    group {
+      id
+    }
+    callQuota
+    visitQuota
+    malanId
+    needsApproval
+    needsMonitor
+  }
```
cainlevy, line 45: do these need to be here? the queries that use MeetingFragment specify `visitors` and `inmates` separately.

src/graphql/fragments/MeetingFragment.graphql
```diff
@@ -0,0 +1,59 @@
...
+  kiosk {
+    id
+    name
+    description
+    enabled
+    group {
+      id
+    }
+    meetingType
+  }
```
cainlevy, line 16: the queries that use MeetingFragment may prefer to decide for themselves if they want `kiosk { ... }`.

src/pages/ContactPage.tsx
```diff
@@ -5,28 +5,35 @@ import { RouteComponentProps } from "react-router";
...
-  const visits = useIPVisits({ contactId: match.params.id });
-  const { loading, error, data } = useQuery(GetVisitorWithRestrictionsDocument, {
+  const { loading, error, data } = useQuery(GetVisitorDocument, {
     variables: { visitorId: match.params.id },
   });
+  const meetingEdges = data?.visitor.meetings.edges
+  const calls: Call[] = []
+  const visits: Visit[] = []
+
+  meetingEdges?.map(edge => {
```
cainlevy, line 33: looks like a `forEach`

src/utils/graphqlConversions.ts
```diff
@@ -0,0 +1,138 @@
...
+import { MeetingType as GraphqlMeetingType } from "src/api/graphql";
+import { MeetingType } from "src/typings/Common";
+import { GetVisitorQuery } from "src/graphql/GetVisitor.generated";
+
+type GraphqlMeeting =
+  | GetInmateQuery["inmate"]["meetings"]["edges"][0]["node"]
+  | GetVisitorQuery["visitor"]["meetings"]["edges"][0]["node"];
+
+const graphQlMeetingStatusToMeetingStatus = (status: GraphqlMeetingStatus) => {
+  const statuses = {
```
cainlevy, line 13: to ensure completeness:

```suggestion
  const statuses: Record<GraphqlMeetingStatus, MeetingStatus> = {
```

src/utils/graphqlConversions.ts
```diff
@@ -0,0 +1,138 @@
...
+    [GraphqlMeetingStatus.Terminated]: "terminated",
+    [GraphqlMeetingStatus.Live]: "live",
+  };
+  return statuses[status] as MeetingStatus;
+};
+
+const graphQlMeetingTypeToMeetingType = (
+  meetingType: GraphqlMeetingType
+): MeetingType => {
+  const types = {
```
cainlevy, line 29:
```suggestion
  const types: Record<GraphqlMeetingType, MeetingType> = {
```

cainlevy (approved):
transitions like this will get messier before they get cleaner. this looks as expected!

src/utils/Message.ts
lspdrz: Ok cool, I'm gonna take care of this in another PR. Only reason I changed here is because it was breaking DOC client

src/graphql/fragments/MeetingFragment.graphql
lspdrz: Good point. Changing

src/graphql/fragments/MeetingFragment.graphql
lspdrz: Agreed

PR #185, opened 2021-12-07
added conditional reason field to connection item

lspdrz:
[sc-1744]

Adds the reason for a rejected connection on the connection item drawer. Note that when we reject a connection anywhere in the app it uses the original nestjs http endpoint, so it won't update the graphql cache and show the reason later on the drawer. The reason will only show on refresh.

image
image
Screen Shot 2021-12-07 at 11 51 09


src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -90,6 +90,8 @@ const getRestrictionDetails = (restriction: ConnectionRestriction) => {
 }
 
 const ConnectionItem: React.FC<Props> = ({ connection, type, onUpdate }) => {
+  console.log('hi')
```
gabrielsaruhashi, line 93: ✂️

gabrielsaruhashi (approved):

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -191,6 +193,10 @@ const ConnectionItem: React.FC<Props> = ({ connection, type, onUpdate }) => {
             <Descriptions.Item label="Status" labelStyle={labelStyle}>
               {`${connection.status}`}
             </Descriptions.Item>
+            {connection.status === "rejected" && 
+            <Descriptions.Item label="Reason" labelStyle={labelStyle}>
+              {`${connection.statusDetails}`}
```
cainlevy, line 198: 🏌️‍♂️ 
```suggestion
              connection.statusDetails
```

PR #190, opened 2021-12-15
read-only view of messages on inmate profile

wittejm:
https://user-images.githubusercontent.com/2104990/146102713-20b26c93-81dd-44b6-8b06-16c4bb39bbfb.mov

This mostly copies the components from pathways on the Messages page. But there's a lot less functionality since it's read-only and staff can see everything. This structure also resembles the components of the MessagesReviewPanel but, again there's a lot less functionality. It results in a fair amount of _kind of but not really_ redundant code. 

Had to fiddle around dealing with the the fact that we're rendering messages at the bottom of the page. I made the messages panel a constant height of `h-96` (24 rems, the max constant height in tailwind). Without this, I first tried bounding the messages list with `max-h-screen`. The result is in the second vid: note how the screen jumps around as the conversation selection changes. 

https://user-images.githubusercontent.com/2104990/146102736-cec6b1a8-a690-40f7-98c5-1fa12e526814.mov


src/components/MessagesReadOnlyPanel/MessageCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { Inmate } from "src/typings/Inmate";
+import { formatDate, genFullName } from "src/utils";
+import { getMessageSenderType, getMessageStatusLabel } from "src/utils/Message";
+import { TConnectionListItem } from "./ConnectionList/ConnectionListItem";
+
+export type TMessageItem =
+  GetConnectionMessagesQuery["connection"]["messages"]["edges"][0]["node"];
+interface Props {
+  message: TMessageItem;
+  connection: TConnectionListItem;
```
gabrielsaruhashi, line 20: oh wait @wittejm 
is the reason why you created a duplicated message card just bc the other lacked the connection?
I'd just add connection to the MessageCard that already existed. The full-fledged version of the implementation also includes the name of the user that sent the message. That was a shortcut I took for the first pass

gabrielsaruhashi: image
i.e you can just expand the functionality of the existing MessageCard

src/components/MessagesReadOnlyPanel/MessageList.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+interface Props {
+  messages: TMessageItem[];
+  connection: TConnectionListItem;
+  className?: string;
+  inmate: Inmate;
+
+}
+
+const MessagesList = ({ messages, connection, inmate }: Props) => {
```
gabrielsaruhashi, line 16: same comment here! You can just reuse the MessageList

src/components/MessagesReadOnlyPanel/ConnectionList/ConnectionList.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const ConnectionList: React.FC<Props> = ({
+  connections,
+  selected,
+  onSelect,
+  className,
+}) => {
+  return (
+    <Layout.Sider
```
gabrielsaruhashi, line 19: nit we can create a HOC/wrapper ConnectionListSidebar? 
This variation that has width 300 and theme light is used in two places

src/components/MessagesReadOnlyPanel/InboxHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { VisitorFragmentFragment } from "src/graphql/fragments/VisitorFragment.generated";
+
+interface Props {
+  visitor: VisitorFragmentFragment;
+}
+
+const InboxHeader = ({ visitor }: Props) => {
+
+  return (
+    <Layout.Header className="bg-white flex items-center justify-between pl-5 pr-5">
```
gabrielsaruhashi, line 14: Let's use a HOC here as well? Seems like it's a copy of the other component

cainlevy: do you mean HOC or component wrapper?

i've mostly seen "hoc" used to describe a function that returns a new component (function or class) each time it is called. that is: a function that returns a function that returns JSX.

wittejm: @gabrielsaruhashi I did copy this but from Pathways. Should this become identical to the header in the Review panel instead? That would add some redundant info:
- The "see connection" link is already available just above in the list of connections. It adds a little convenience to make it clickable here too, but it might be a little crowding with
- The InmateName & VisitorName pattern is has redundancy with the fact that this panel is appearing on the particular inmate's page. Include it anyway?

wittejm:
Cool, reusing MessageCard and MessageList was simpler than I had assumed. The next three commits extract out wrapper components for the Sidebar and Header -- There's only a little bit of replicated style to pull out.

src/components/MessagesReadOnlyPanel/InboxHeader.tsx
wittejm: Answered this question in my new commit that still replicates the style from the Pathways view... see below.

src/graphql/fragments/MessageFragment.graphql
```diff
@@ -8,7 +8,12 @@ fragment MessageFragment on Message {
   createdAt
   availableAt
   sender {
-    id
+    ... on Inmate {
```
wittejm, line 11: @tanyabrassie These inline fragments fix that error you're seeing

src/components/MessageReviewPanel/MessagesListHeader.tsx
```diff
@@ -0,0 +1,17 @@
+import { Layout } from "antd";
+import React, { ReactElement } from "react";
+
+interface Props {
+  content: ReactElement
```
gabrielsaruhashi, line 5: have you used the special props "children" in the past?
if you type it as React.FC<props> you'll automatically gain access to it.
It consists of a an array of React.ReactNode || React.ReactNode[]

src/components/MessageReviewPanel/MessagesListHeader.tsx
```diff
@@ -0,0 +1,17 @@
...
+  className?: string
+}
+
+const MessagesListHeader = ({ content, className }: Props) => {
```
gabrielsaruhashi, line 9: const MessagesListHeader: React.FC<Props> = ({ children, className})

src/components/MessageReviewPanel/MessagesListHeader.tsx
gabrielsaruhashi: i.e you don't need to define this "content" props here if you use the default 'children'

src/components/MessageReviewPanel/ConnectionListSidebar/ConnectionListSidebar.tsx
```diff
@@ -1,59 +1,20 @@
...
-};
+import { Layout } from "antd";
+import React, { ReactElement } from "react";
 
 interface Props {
-  flaggedMessageMap: ConnectionFlaggedMessagesMap;
-  selectedConnection: TConnectionListItem | undefined;
-  onSelectConnection: (connection: TConnectionListItem) => void;
-  className?: string;
+  connectionList: ReactElement,
```
gabrielsaruhashi, line 5: same here! you can cut this out and use the 'children' keyword

src/components/MessageReviewPanel/ConnectionListSidebar/ConnectionList.tsx
```diff
@@ -0,0 +1,56 @@
...
+  onSelectConnection: (connection: TConnectionListItem) => void;
+  className?: string;
+}
+
+const ConnectionList = ({
+  flaggedMessageMap,
+  onSelectConnection,
+  selectedConnection,
+}: Props) => {
+  const uniqueConnections = Object.values(flaggedMessageMap).map(
```
gabrielsaruhashi, line 25: i might have written this logic so it's my bad for not remembering lol
but how is this ensuring uniqueness?

src/components/MessageReviewPanel/MessagesListHeaderContent.tsx
```diff
@@ -0,0 +1,24 @@
...
+const MessagesListHeaderContent = ({ connection }: Props) => {
+
+  return (
+    <>
+      <span className="font-bold text-lg">
+        {genFullName(connection.inmate)} {"&"}{" "}
+        {genFullName(connection.visitor)}{" "}
+      </span>
+      {/* TODO: placeholder */}
+      <Typography.Link>See connection</Typography.Link>
```
gabrielsaruhashi, line 19: Oh @lspdrz did implement this I believe
@wittejm want to add that logic in? could be a separate PR if you prefer

src/components/MessageReviewPanel/MessagesReviewPanel.tsx
```diff
@@ -36,22 +38,19 @@ const MessagesReviewPanel = ({
...
         <Layout>
-          <Layout.Header className="bg-white flex items-center gap-2">
-            <span className="font-bold text-lg">
-              {genFullName(selected.connection.inmate)} {"&"}{" "}
-              {genFullName(selected.connection.visitor)}{" "}
-            </span>
-            {/* TODO: placeholder */}
-            <Typography.Link>See connection</Typography.Link>
-          </Layout.Header>
+          <MessagesListHeader className="gap-2" content={
```
gabrielsaruhashi, line 49: if you start using children you can just do 
<MessageListHeader>
  <MessageListHeaderContent />
</MessageListHeader>

gabrielsaruhashi: this effectively becomes a general purpose wrapper

gabrielsaruhashi (approved):
lgtm! 
just take a look at the children suggestion ? and let me know if u any questions

src/components/MessageReviewPanel/ConnectionListSidebar/ConnectionList.tsx
wittejm: Not enforced here, but when the map is built here: https://github.com/Ameelio/connect-doc-client/blob/932c5ad1bae86e88a815c10054d9aeb4e236336b/src/pages/MessageReviewPage.tsx#L46

src/components/MessageReviewPanel/MessagesListHeaderContent.tsx
wittejm: Gave myself another ticket!

PR #192, opened 2021-12-16
display when search meetings pages are loading

cainlevy:
1. Connects the new GraphQL query's `loading` indicator through to the `<SearchMeetingsTable />` component.
2. Removes the Redux query that was no longer used

Search remains broken for ICIW because the date range will not affect which 500 meetings can be filtered.


src/components/SearchMeetings/SearchMeetings.tsx
```diff
@@ -67,28 +59,6 @@ const SearchMeetings = ({
...
-      }))().then(() => setLoading(false));
-  }, [
-    fetchMeetings,
-    limit,
-    offset,
-    startDate,
-    endDate,
-    global,
-    activeSearchFilter,
-  ]);
```
cainlevy, line 90: it's hard to remove redux calls confidently because they have side effects on distant places via selectors, but:

1. i tried enabling and disabling this fetch and didn't notice any effect
2. i verified that the filtered meetings used the graphql data provided as a prop
3. i couldn't find any redux selectors

gabrielsaruhashi: there shouldn't be side effects for this one!

gabrielsaruhashi: at least negative ones

src/pages/SearchInPersonVisitsPage.tsx
```diff
@@ -1,44 +1,34 @@
...
-  const fetchMeetings = useCallback(
-    (filters: MeetingFilters) => {
-      dispatch(fetchInPersonVisits(filters));
-    },
-    [dispatch]
-  );
+  const visits = data
+    ? data.facility.meetings.edges
+        .filter((e) => e.node.meetingType !== MeetingType.Call)
+        .map((e) => graphQlMeetingToVisit(e.node))
```
gabrielsaruhashi, line 20: this is a cool util

cainlevy: yeah! @lspdrz added these so we can start transitioning.

src/pages/SearchCallsPage.tsx
```diff
@@ -1,45 +1,34 @@
...
 import { graphQlMeetingToCall } from "src/utils/graphqlConversions";
 
-
 const SearchCallsPage = () => {
-  const dispatch = useAppDispatch();
   const facility = useAppSelector(
     (state: RootState) => state.facilities.selected
   );
 
   const { loading, error, data } = useQuery(GetFacilityMeetingsDocument, {
```
gabrielsaruhashi, line 14: the change on the API side was to limit to 500? / 
or is there any logic rn that passes the date range as well

gabrielsaruhashi (approved):

src/pages/SearchCallsPage.tsx
cainlevy: yes, the API limits to 500.

the client could ask for more or less, but there's no option for date filtering (yet). we need to decide how we're going to fix this search.

PR #194, opened 2021-12-17
Lp call messages gql

lspdrz:
[sc-1314]

DO NOT MERGE UNTIL https://github.com/Ameelio/connect-api-nest/pull/582 is merged and deployed

This page isn't used anywhere so there's not really a way to test


src/graphql/fragments/CallMessageFragment.graphql
```diff
@@ -0,0 +1,8 @@
+fragment CallMessageFragment on CallMessage {
+  id
+  createdAt
+  sender {
```
wittejm, line 4: I've seen this give an error before if it doesn't include the in-line fragment structure like this:
https://github.com/Ameelio/pathways-client/blob/089e8e9a80970bd40253bb71b97ababee7e5d243/src/graphql/queries/GetConnectionMessages.graphql#L12

I'm surprised if this is working as-is.

wittejm (approved):

src/graphql/fragments/CallMessageFragment.graphql
lspdrz: Weird. I'm gonna change it to that anyways

PR #196, opened 2021-12-17
search meetings by type

cainlevy:
The GraphQL endpoint currently returns only the last 500 meetings. We can slightly improve the usefulness of these 500 results by asking GraphQL to filter by type.

I reconsidered additionally filtering by meeting status. I don't think we should let users add query filters until we can limit the result size to something more reasonable. That probably means going all in on search filters and pagination. Or cutting scope on this page. cc @gabrielsaruhashi


lspdrz (approved):
Great idea

PR #199, opened 2021-12-22
changes fetchFacilities function to use gql

lspdrz:
[sc-1319]

DO NOT MERGE UNTIL https://github.com/Ameelio/connect-api-nest/pull/590 is merged and deployed


src/api/client.ts
```diff
@@ -20,3 +21,13 @@ export const client = new ApolloClient({
   cache: new InMemoryCache(),
 });
 
+export const urqlClient = createClient({
```
gabrielsaruhashi, line 24: wait why are we using urql?

lspdrz: I was getting type errors with the apollo client, and this is the package that Lance uses for the gql calls in the redux files in pathways

gabrielsaruhashi: oh i see - is there a way to use the apollo client we set up to do this call on the redux store?

lspdrz: I have no idea, but I can explore

src/redux/modules/facility.ts
```diff
@@ -45,15 +47,15 @@ export const selectActiveFacility = createAsyncThunk(
 export const fetchFacilities = createAsyncThunk(
   "facility/fetchFacilities",
   async () => {
-    const fBody = await fetchAuthenticated(
-      `users/${Store.getState().session.user.id}/facilities`,
-      {},
-      false
-    );
+    const result = await urqlClient
```
gabrielsaruhashi, line 50: what's the next step for this? 
I imagine we want to move this to a React Context Provider

lspdrz: I'm not too sure. I've found what you and Lance mention about moving away from redux sort of ambiguous tbh, but I haven't questioned y'all about it. Are we moving to react context? What for, as opposed to redux?

Concerning the above, I think the main move here is to simply create the GQL query and get the API up to date with the fields we need. That seems to be the sort of intermediary steps Lance is taking in pathways. And then after, I think we can move the queries to more appropriate places as you say, once we make decisions on what we're gonna use for state management

wittejm (approved):

PR #201, opened 2021-12-24
DOC Voice Calls Scaffolding 

gabrielsaruhashi:
image

### Description 
- feature flag 
- utils to transform from graphql meeting type to voice call type
- extracted hook to handle meeting type 

### Dependencies
Depends on https://github.com/Ameelio/connect-api-nest/pull/595

[sc-2031]
[sc-2032]


src/hooks/useMeetingsByType.ts
```diff
@@ -0,0 +1,50 @@
...
+  graphQlMeetingToVisit,
+  graphQlMeetingToVoiceCall,
+} from "../utils/graphqlConversions";
+
+export default function useMeetingsByType(meetings: GraphqlMeeting[]): {
+  voiceCalls: VoiceCall[];
+  videoCalls: Call[];
+  visits: Visit[];
+} {
+  const [videoCalls, setVideoCalls] = useState<Call[]>([]);
```
gabrielsaruhashi, line 16: before, this was being re-computed and led to multiple re-renders of the search meetings table component

src/typings/Call.ts
```diff
@@ -114,7 +120,9 @@ export type Visit = BaseInPersonVisit & MeetingDetails;
 
 export type Call = BaseCall & MeetingDetails;
 
-export type Meeting = Visit | Call;
+export type VoiceCall = BaseVoiceCall & MeetingDetails;
```
gabrielsaruhashi, line 123: we're very close to be able to nix these
I might take a stab at it sometime soon
For now, I'll keep the pattern!

wittejm (approved):

src/hooks/useMeetingsByType.ts
cainlevy: this sorts meetings by type, yeah?

could be a single `useMemo` (`useEffect` + `useState` = `useMemo`), or a plain transform function that ContactPage and InmatePage include in the `useMemo` hooks they currently have.

cainlevy (approved):

src/components/Profile/ProfileMeetingsCard.tsx
```diff
@@ -1,17 +1,44 @@
...
+  {
+    key: "voice_calls",
+    tab: "Voice Calls",
+  },
+];
+
+const ProfileMeetingsCard = ({ videoCalls, visits, voiceCalls }: Props) => {
   const [activeMeetingTab, setActiveMeetingTab] = useState(MEETING_TABS[0]);
 
+  const activeMeetings = useMemo(() => {
```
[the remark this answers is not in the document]
gabrielsaruhashi: I agree it's slower!
But this will be recomputed every time if we don't wrap it in a hook, won't it? I assumed it would lead to re-render as well in this case.
Will the react component recognize it's the same array? Seems analogous to the case we transform the array with a .map/.filter
I'll keep it like this for now, and experiment when Im back to learn - v curious to learn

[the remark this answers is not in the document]
cainlevy: 🤔 well that's a puzzle. i didn't know javascript well enough so i designed a little experiment and confirmed that multiple references to the same object are considered strictly equal:

```
const myObj = {foo: 'bar'};
const other = {foo: 'bar'};

// objects have equal content but are not strictly equal
console.log(myObj === other); // false

// however two references to the same object are strictly equal
const one = myObj;
const two = myObj;
console.log(one === two); // true
```

so i would not expect re-renders in this case.

that also fits with the fact that even in the `useMemo` scenario we're assigning the variable on each render. it works because the value remains strictly equal.

PR #203, opened 2021-12-24
create live voice call monitoring page

gabrielsaruhashi:
Up next:

image


src/hooks/useSelectedFacility.ts
```diff
@@ -0,0 +1,9 @@
+import { useAppSelector } from "src/redux";
+import { SelectedFacility } from "src/typings/Facility";
+
+export default function useSelectedFacility(): SelectedFacility {
+    const selected = useAppSelector(state => state.facilities.selected);
+    if (!selected) throw new Error('Facility must be selected');
```
wittejm, line 6: When can this error throw in practice?

cainlevy (approved):
my memory of the product meeting is that we had some unanswered questions about how live call monitoring would work, especially with respect to pagination. i left with the impression that this page's design was unfinished.

gabrielsaruhashi:
@cainlevy 
> my memory of the product meeting is that we had some unanswered questions about how live call monitoring would work, especially with respect to pagination. i left with the impression that this page's design was unfinished.

I see. Based on the meeting notes at the bottom of the doc, I think the question was re: whether the user could listen to multiple calls at once, and we talked about a spotify player-like UI/UX that only allowed the DoC to be listening to one call at a time.  That seemed to solve the concern re: pagination as the DoC would not have several active connections at a time.

Voice calls are treated differently than video as video = visitation and DoCs felt it was imperative to be able to watch all visitations at once. For voice calls, we'll have more leeway to design how the experience looks like.

src/pages/LiveVoiceCallMonitoringPage.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const { data, loading } = useQuery(GetFacilityMeetingsDocument, {
+    variables: {
+      id: facility.id,
+      meetingType: [MeetingType.VoiceCall],
+      meetingStatus: [MeetingStatus.Live],
+    },
+  });
+
+  const voiceCalls = useMemo(
+    () => data?.facility.meetings.edges.map(({ node }) => node).map(graphQlMeetingToVoiceCall) || [],
```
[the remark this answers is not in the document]
gabrielsaruhashi: yep! for now. Planning to refactor when I'm back :) 
Will open a ticket

gabrielsaruhashi:
@cainlevy merging this in! but the discussion re: whether listening into multiple voice calls is needed / not needed is not fully validated yet.
We can start by assuming a "no" as that reduces complexity by a lot

src/hooks/useSelectedFacility.ts
gabrielsaruhashi: not really anywhere currently.
We should only let users access internal pages when they are logged in (which requires a selected facility)
We just had this type narrowing logic across all pages before so I'm centralizing in this hook in case that ever changes

PR #204, opened 2022-01-04
show incarcerated individual's location

cainlevy:
Previously the "Location" cell was empty. This made it more difficult to debug a DWCF onboarding issue.

image


src/pages/InmatesPage.tsx
```diff
@@ -65,53 +70,6 @@ const UnconnectedInmateContainer: React.FC<PropsFromRedux> = ({
...
-    },
-    {
-      title: "Last Name",
-      dataIndex: "lastName",
-      editable: true,
-    },
-    {
-      title: "Location",
-      dataIndex: "location",
-      editable: true,
```
wittejm, line 96: Why were these editable: true before and why is that gone now?

wittejm (approved):

src/pages/InmatesPage.tsx
cainlevy: i think it was incomplete support for editable table cells? it didn't type-check when i moved the configuration into antd components.

PR #205, opened 2022-01-05
Use Graphql for Connection fetch and update on Contact/Inmate Pages

lspdrz:
DO NOT MERGE UNTIL https://github.com/Ameelio/connect-api-nest/pull/598 is merged and deployed

[sc-1933]
[sc-1315]

https://user-images.githubusercontent.com/9705113/148221124-a8251e49-a313-49ab-ad0d-a955dd87adeb.mov


lspdrz:
Video is slightly inaccurate, I changed the status field on the drawer to title case 
image

src/components/Profile/Connections/ConnectionItem.tsx
```diff
@@ -1,23 +1,23 @@
...
 }
 
-const ConnectionItem: React.FC<Props> = ({ connection, type }) => {
+type Connection =
+  | GetVisitorQuery["visitor"]["connections"][0]
+  | GetInmateQuery["inmate"]["connections"][0];
 
-  const person = type === "inmate" ? connection.contact : connection.inmate;
+const ConnectionItem: React.FC<Props> = ({ connection, type }) => {
+  const person = type === "inmate" ? connection.visitor : connection.inmate;
```
gabrielsaruhashi, line 20: I feel we're doing this in a couple of places! Wonder if we want to extract it as a util.
Correct me if I'm wrong tho !

src/pages/ContactPage.tsx
```diff
@@ -47,23 +49,24 @@ function ContactPage({ match }: RouteComponentProps<TParams>): ReactElement {
     dispatch(fetchContactIdImages(contact.id));
   }, []);
 
-  const connections = useContactConnections(contact?.id || "");
+  const connections = data?.visitor.connections;
 
-  if (!contact || !facilityName) return <div />;
+  if (!contact || !facilityName || !connections) return <div />;
```
gabrielsaruhashi, line 54: should we show a loading spinner? 
Check "PageLoader" we might want to tweak that as we haven't used it in a while

src/pages/InmatePage.tsx
```diff
@@ -33,21 +35,22 @@ function InmatePage({ match }: RouteComponentProps<TParams>): ReactElement {
...
+  const visits: Visit[] = [];
 
   // Temporary solution to order meetings from most recent to oldest. We have to spread first because the original array is read only
-  const reversedMeetingEdges = meetingEdges && [...meetingEdges].reverse()
+  const reversedMeetingEdges = meetingEdges && [...meetingEdges].reverse();
 
-  reversedMeetingEdges?.forEach(edge => {
-    edge.node.call ? calls.push(graphQlMeetingToCall(edge.node)) : visits.push(graphQlMeetingToVisit(edge.node))
-  })
+  reversedMeetingEdges?.forEach((edge) => {
```
gabrielsaruhashi, line 49: @lspdrz check the useMeetingsByType hook that i just created?

It will probably give you a merge conflict but it will prevent you from doing multiple re-rerenders of this page

src/graphql/fragments/ConnectionFragment.graphql
```diff
@@ -1,4 +1,46 @@
 fragment ConnectionFragment on Connection {
   id
   status
+  relationship
+  inmate {
```
wittejm, line 5: I think fragments shouldn't contain child objects, only primitives, to increase reusability of the fragments. So I prefer the previous approach that was using additional fragments on child objects. Benefits:
- We won't over-fetch data on Connection when we don't need it for a different query but still want to use a connection fragment.
- We don't have redundant fragments (i.e., the connection fragment and inmate fragment won't have all the same primitive fields)
-

src/pages/InmatePage.tsx
lspdrz: i merged in your changes!

src/components/Profile/Connections/ConnectionItem.tsx
lspdrz: the only place i can see is in videohandler, I think we're ok keeping it as it is now

lspdrz: Yeah, we are doing it in some places. If we want to refactor, I'd suggest making a SC ticket and doing in a new branch

src/pages/ContactPage.tsx
lspdrz: Sure thing it's done

src/graphql/fragments/ConnectionFragment.graphql
lspdrz: @wittejm Agreed on all, moved the objects back to the queries

src/graphql/GetInmate.graphql
```diff
@@ -2,11 +2,47 @@ query GetInmate($inmateId: ID!) {
   inmate(id: $inmateId) {
     ...InmateFragment
     connections {
+      ...ConnectionFragment
+      inmate {
+        id
```
wittejm, line 7: These fields can all be replaced with the InmateFragment right? Sorry if my last comment was unclear; I meant that we can define the set of primitive fields on an object once (in its fragment) and then reuse that everywhere. In this case the facility and group fields would stick around because they're objects and we should remove those from the InmateFragment.

Same deal with VisitorFragment here, and in the GetVisitor query below. Does Apollo complain about using the same fragment in a query twice? (that would be lame!)

lspdrz: I think it does, I think I tried that before. I'm confused on what you mean here, do you want to edit it directly and show me?

wittejm: Sure!

wittejm (approved):

PR #208, opened 2022-01-06
tailwind overrides antd

gabrielsaruhashi:
We want the custom tailwind styling to take precedence over antd defaults


wittejm (approved):

wittejm:
:raised_hands:

PR #209, opened 2022-01-06
changed inmatespage and inmatepage to use gql

lspdrz:
[sc-1322]

DO NOT MERGE until https://github.com/Ameelio/connect-api-nest/pull/672 is merged and deployed


gabrielsaruhashi:
@lspdrz as we discussed before, i dont think we need pagination for inmates.
i'd wait until it becomes a real problem that we need to solve

lspdrz:
> @lspdrz as we discussed before, i dont think we need pagination for inmates. i'd wait until it becomes a real problem that we need to solve

@gabrielsaruhashi sorry i don't follow? This isn't for pagination, this is simply to switch the fetchInmates query to gql

gabrielsaruhashi:
@lspdrz your note is saying:
>DO NOT MERGE. In Draft Mode until we figure out the pagination for inmates

lspdrz:
> @lspdrz your note is saying:
> 
> > DO NOT MERGE. In Draft Mode until we figure out the pagination for inmates

@gabrielsaruhashi Ah right, that was actually referring to a bigger conversation I was having with @cainlevy . We were getting performance issues with loading a lot of records at once for /meetings and I didn't want to introduce here, so we paused the implementation until we figure out a bigger solution.

I think the band-aid solution right now for meetings is to cap at 500, which we can't do for inmates I think bc we have facilities with >500 inmates. So some won't be loaded if we go that route

gabrielsaruhashi:
@lspdrz i dont see how this concern is blocking. we're already doing that for the CRUD API endpoint.

afaik, the reason why it's extra expensive on meetings is that there are several joins to load a meeting entity.
we should move forward with this and just return all inmates at once. we want to finish refactoring asap.

cainlevy:
the REST endpoint may fetch all inmates, but the GraphQL endpoint would fetch all inmates plus associated data. i can't guess what impact that would have.

when i noticed the performance problem on meetings, i added a server-side limit: 500 failsafe on all relay connections. we can override it for `FacilityResolver.inmates` if we want to find out what happens.

long term solutions might look like:

- figure out client side pagination ([start here](https://www.apollographql.com/docs/react/pagination/cursor-based/))
- consider a "only the first 100 are loaded, please refine your search" kind of interaction

gabrielsaruhashi:
@lspdrz  could you share the results of this performance problem you're alluding to? How long does it currently take to load 1,000 inc people? Avg prison facility in the US houses 500, it will take a while for us to hit a perf problem - I'd refrain from overengineering.

This will be a query with plain InmateFragments, that's why I don't think it will be as slow as the meetings query.

lspdrz:
@gabrielsaruhashi I don't know what the performance problems are, the relevant part is that the solution put in place for the performance problems on the meetings query was a default 500 server-side limit. This default limit stalled the PR until further discussion regarding the above solutions that Lance mentioned:
- override for inmates query
- figure out client side pagination (start here)
- consider a "only the first 100 are loaded, please refine your search" kind of interaction

We're not overengineering here, if we merge and deploy the code without selecting one of the above solutions than only a subset of the inmates at a facility would be returned in a query, namely the first 500 retrieved

gabrielsaruhashi:
@lspdrz going back to the original discussion and the point I am trying to make since the start of the task:
https://ameelioorg.slack.com/archives/C01AG36A0P6/p1641476616004400?thread_ts=1641475924.003400&cid=C01AG36A0P6

I don't think we need to paginate this query and I am proposing we try the simple solution instead of going for the complex one. Right now we're using InmateConnection, and that's why the relay limits are blocking us. The connection concept is used for pagination. If you used plain queries this wouldn't be a blocker as there's no relay.

We'll definitely hit that wall eventually (e.g when we onboard Riker), but is this the right time to change this page tho given everything else we have going on? That's why I am asking whether the performance hit is actually concerning.

lspdrz:
@gabrielsaruhashi cool sounds good to me. I'm making a note to revisit this tomorrow morning and change the inmate query to stop using InmateConnection

wittejm (approved):

PR #210, opened 2022-01-11
Emessaging CSV

lspdrz:
DO NOT MERGE UNTIL https://github.com/Ameelio/connect-api-nest/pull/618 is merged and deployed

[sc-1494]

Demo:

https://user-images.githubusercontent.com/9705113/149015246-ec10a2c1-e313-4818-b982-81b3debc8ff5.mov


src/graphql/fragments/MessageFragment.graphql
```diff
@@ -3,17 +3,32 @@ fragment MessageFragment on Message {
...
   }
   status
+  statusDetails
+  internalComments
   createdAt
   availableAt
   sender {
     ... on Inmate {
       id
+      inmateIdentification
```
wittejm, line 19: I think since the set of fetched data fields on Inmate and Visitor are getting larger in this query, we should pull them out of the fragment and define them in the top-level graphql query instead. Same idea as what I recommended on your PR last week.

src/graphql/fragments/MessageFragment.graphql
```diff
@@ -3,17 +3,32 @@ fragment MessageFragment on Message {
...
     }
     ... on Visitor {
       id
+      firstName
+      lastName
     }
   }
+  reviewedAt
+  reviewedBy {
+    ...StaffFragment
```
wittejm, line 31: I've gotten an error (a runtime error, I think?) from using a fragment in a fragment before, it's weird that I did but you don't get an error here. 

But I think it's a better idea to not nest fragments anyway: it makes the fragments more minimal so that we don't over-fetch child objects in other queries when we don't need them, and it makes the top-level graphql queries more readable since we can directly see in those queries which child objects are getting fetched or not.

wittejm:
In the worse case with nesting fragments, we might create a symmetric recursion because we want, e.g.,  Inmate with its child Messages in one fragment and Message with its child Inmate in another. Maybe that's the error I was getting. But having nested fragments at all is the hazard that would invite that problem, IMO.

src/graphql/fragments/MessageFragment.graphql
gabrielsaruhashi: @lspdrz yes, we shouldnt nest fragments
I was doing that on the my very first pass on the clients, and Lance advised against it bc of these loopholes

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
```diff
@@ -1,62 +1,144 @@
...
-                  <MessagesListHeader className="justify-between pl-5 pr-5">
-                    <MessagesListHeaderContent visitor={selected.visitor} />
-                  </MessagesListHeader>
-                  <MessagesList messages={messages} />
-                  <Layout.Footer />
-                </>)
-              }
+        <Card title={"Messages"}>
+          <div className="flex flex-col justify-between">
+            <div className="flex justify-between mb-5">
```
gabrielsaruhashi, line 80: Let's move this to its own component? MessageCsvExport or smt
There's a lot of logic that was introduced just for this export feature and it's making it hard to read the Panel

lspdrz:
@wittejm @gabrielsaruhashi all changes done

gabrielsaruhashi:
@lspdrz dont forget to re-request reviews so we get notifs on our inboxes!

src/components/MessagesReadOnlyPanel/MessagesCSVExport.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+interface Props {
+  messages: Message[];
+  inmate: Inmate;
+  visitor: VisitorFragmentFragment;
+  startDate: number;
+  endDate: number;
+}
+
+const MessagesCSVExport = ({
```
gabrielsaruhashi, line 21: much cleaner! 🧼

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
```diff
@@ -1,62 +1,108 @@
...
+      let start = new Date(startDate);
+      let end = new Date(endDate);
+      filtered = filtered.filter(
+        (message) =>
+          new Date(message.createdAt) >= start &&
+          new Date(message.createdAt) <= end
+      );
+    }
+    setFilteredMessages(filtered);
+  }, [messages, setFilteredMessages, startDate, endDate]);
```
gabrielsaruhashi, line 56: u sure setFilteredMessages is needed here?

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
```diff
@@ -1,62 +1,108 @@
...
-  const messages = (messagesData?.connection.messages.edges.map((e) => e.node)) || [];
+  const messages = useMemo(
+    () => messagesData?.connection.messages.edges.map((e) => e.node) || [],
+    [messagesData]
+  );
+  useEffect(() => {
+    if (!messages.length) return;
+    let filtered = messages;
+    if (startDate && endDate) {
+      let start = new Date(startDate);
```
gabrielsaruhashi, line 47: const start

src/utils/Message.ts
```diff
@@ -1,5 +1,81 @@
...
+    }
+    if (message.statusDetails) {
+      details += `\nMessage to sender: ${message.statusDetails}`;
+    }
+    if (message.internalComments) {
+      details += `\nInternal comments: ${message.internalComments}`;
+    }
+
+    return {
+      ["Timestamp"]: timestamp,
```
gabrielsaruhashi, line 73: seems like there are 3 noticeable blocks of information

could we move each block to its own private implementation to make this more readable and modular

i.e formatMessageTimestamp
formatMessageContent
formatMessagedetails

src/utils/Message.ts
```diff
@@ -1,5 +1,81 @@
...
+    if (message.availableAt) {
+      timestamp += `\nDelivered ${format(
+        new Date(message.availableAt),
+        "MM/dd/yy H:mm"
+      )}`;
+    } else {
+      timestamp += `\nNot delivered`;
+    }
+
+    const sender = message.sender as any;
```
gabrielsaruhashi, line 26: why do we need to force cast this to any 🧐 ?
Seems like you are assuming this is of instance Inmate

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
```diff
@@ -1,62 +1,108 @@
...
+                  onStartDateChange={setStartDate}
+                  onEndDateChange={setEndDate}
+                />
+                <MessagesCSVExport
+                  messages={filteredMessages}
+                  inmate={inmate}
+                  visitor={selected.visitor}
+                  startDate={startDate ? startDate : messages[0].createdAt}
+                  endDate={
+                    endDate ? endDate : messages[messages.length - 1].createdAt
```
gabrielsaruhashi, line 75: shouldnt this be 

endDate ? endDate : new Date()

Even if the last message was sent a month ago, we should be stating that the report captures all messages that were sent to this current date

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
```diff
@@ -1,62 +1,108 @@
...
+              <div className="flex justify-between mb-5">
+                <MessagesFilter
+                  onStartDateChange={setStartDate}
+                  onEndDateChange={setEndDate}
+                />
+                <MessagesCSVExport
+                  messages={filteredMessages}
+                  inmate={inmate}
+                  visitor={selected.visitor}
+                  startDate={startDate ? startDate : messages[0].createdAt}
```
gabrielsaruhashi, line 73: same comment here-  i think it this gives the misleading impression that we're only getting a snapshot of the message history. Also I believe this will throw an error when there are no messages 

we should just do

startDate={startDate}

and adapt the copy on messageCsvExport that says that these are all the messages

src/components/MessagesReadOnlyPanel/MessagesCSVExport.tsx
```diff
@@ -0,0 +1,40 @@
...
+  inmate,
+  visitor,
+  startDate,
+  endDate,
+}: Props) => {
+  const fileNameStart = `${inmate.lastName}, ${inmate.firstName} ${inmate.inmateIdentification} <>  ${visitor?.lastName}, ${visitor?.firstName} ${visitor?.email}`;
+
+  const firstDate = format(startDate, "yyyy-MM-dd");
+  const lastDate = format(endDate, "yyyy-MM-dd");
+  const filename = `${fileNameStart} [${firstDate} - ${lastDate}].csv`;
```
gabrielsaruhashi, line 32: if firstDate is nullable it means that this captures ALL of the messages that were ever sent
Copy needs to be updated here.

I'd change it to:

`${fileNameStart} [${lastDate}].csv`

if there's no start date

src/components/MessagesReadOnlyPanel/MessagesCSVExport.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { messagesToCsv } from "src/utils/Message";
+
+type Message =
+  GetConnectionMessagesQuery["connection"]["messages"]["edges"][0]["node"];
+
+interface Props {
+  messages: Message[];
+  inmate: Inmate;
+  visitor: VisitorFragmentFragment;
+  startDate: number;
```
gabrielsaruhashi, line 17: should be nullable here

src/utils/Message.ts
lspdrz: Because inmateIdentification doesn't exist on Visitor and I need to test against it

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
lspdrz: But startDate is possibly null

src/components/MessagesReadOnlyPanel/MessagesCSVExport.tsx
lspdrz: This goes against Liz's design , can you confirm with her?

src/components/MessagesReadOnlyPanel/MessagesCSVExport.tsx
lspdrz: Why should this be nullable?

src/utils/Message.ts
cainlevy: should be able to narrow types with `message.sender.__typename`

src/components/MessagesReadOnlyPanel/MessagesFilter.tsx
```diff
@@ -0,0 +1,43 @@
...
+          onEndDateChange(null);
+          return;
+        }
+        if (date[0]) {
+          const startEpoch = date[0].unix() * 1000;
+          const startDate = new Date(startEpoch);
+
+          startDate.setHours(0);
+          startDate.setMinutes(0);
+          startDate.setSeconds(0);
```
cainlevy, line 27: this assumes the user's time zone will be the same as the facility. if not, the "beginning of day" and "end of day" calculations can be some hours off.

fixing would probably involve sending date strings like `YYYY-MM-DD` to the backend where they can be interpreted in the facility's time zone. ugh. may not be worth it?

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
```diff
@@ -1,62 +1,107 @@
...
     variables: { id: selected?.id || "" },
     skip: !selected?.id,
   });
-  const messages = (messagesData?.connection.messages.edges.map((e) => e.node)) || [];
+  const messages = useMemo(
+    () => messagesData?.connection.messages.edges.map((e) => e.node) || [],
+    [messagesData]
+  );
+  useEffect(() => {
+    setFilteredMessages(messages);
```
cainlevy, line 44: i recommend `useMemo` for effect+state patterns like this

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
```diff
@@ -1,62 +1,107 @@
...
+    if (!messages.length) return;
+    let filtered = messages;
+    if (startDate && endDate) {
+      const start = new Date(startDate);
+      const end = new Date(endDate);
+      filtered = filtered.filter(
+        (message) =>
+          new Date(message.createdAt) >= start &&
+          new Date(message.createdAt) <= end
+      );
```
cainlevy, line 54: should we filter on endDate even if there's no startDate?

cainlevy (approved):

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
gabrielsaruhashi: exactly. but messages[0].createdAt is also null (and in this case throws errors)
unless you're ensuring that the array length >= 1

lspdrz: So startDate will never possibly be null

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
lspdrz: Interesting usecase. And maybe the same for startDate if there is no endDate? Antd filter doesn't let us create these scenarios though so I think we're in the clear

PR #217, opened 2022-01-18
added polling for messages

lspdrz:
[sc-2312]


src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
```diff
@@ -29,10 +30,17 @@ const MessagesReadOnlyPanel = ({ inmate }: Props) => {
...
+    GetConnectionMessagesDocument,
+    {
+      variables: { id: selected?.id || "" },
+      skip: !selected?.id,
+      pollInterval: 30000,
+    }
+  );
+
+  if (loading)
+    showToast("loading", "Fetching the latest updates...", "loading");
```
cainlevy, line 43: i would suggest `useEffect` but the toast key will probably dedupe well enough

cainlevy (approved):

src/components/MessagesReadOnlyPanel/MessagesReadOnlyPanel.tsx
gabrielsaruhashi: @lspdrz can we create a more specific key to avoid any accidental naming collision? 
"loading/messagesReadOnlyPanel"

PR #218, opened 2022-01-20
doc admins can view the profile of a staff + graphql refactor for staff page

gabrielsaruhashi:
[sc-1325]

https://user-images.githubusercontent.com/12942772/150369509-779e23e3-ddd7-4bb8-8a74-ea86147d2141.mov

### Up Next
Gonna do a quick lunch break now, but I just need to add the "Affected Parties" column that links the profiles of the users that were affected by the action. I'll also create a reusable link to staff profiles as part of this work :)


src/components/Profile/ProfileHeader/StaffProfileHeader.tsx
```diff
@@ -0,0 +1,61 @@
...
+    {
+      path: "/",
+      breadcrumbName: facility.name,
+    },
+    {
+      path: "/staff",
+      breadcrumbName: "Staff",
+    },
+    {
+      path: "/",
```
cainlevy, line 30: what's our convention for the current entry in breadcrumbs? is it a link?

src/pages/StaffRosterPage.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const facility = useSelectedFacility();
+  const dispatch = useAppDispatch();
+  const [removalSelection, setRemovalSelection] = useState<StaffPosition>();
+  const [confirmLoading, setConfirmLoading] = useState(false);
+  const isUserAdmin = user.staffPositions.some(
+    (position) =>
+      facility &&
+      position.facilityId === facility.id &&
+      position.role === "admin"
+  );
```
cainlevy, line 36: i'll bet this turns into a common hook, e.g. `useSelectedFacilityRole`

cainlevy (approved):

src/pages/StaffRosterPage.tsx
gabrielsaruhashi: i think tony created one actually! 
i just copy pasted this piece. let me check

src/components/Profile/ProfileHeader/StaffProfileHeader.tsx
gabrielsaruhashi: oh oop ! forgot to correct this.
Yes it should be - although our current implementation of the hash router (with redux) is not happy with antd's breadcrumbs :x 
It's been in the back-burner for a while, might take a look soon

PR #220, opened 2022-01-22
graphql version of pending connection requests

gabrielsaruhashi:
Depends on https://github.com/Ameelio/connect-api-nest/pull/655

https://user-images.githubusercontent.com/12942772/150654069-77c54050-47e5-4862-b4bf-f9f623430c7b.mov


src/utils/Connection.ts
```diff
@@ -36,3 +40,49 @@ export function loadAllConnectionEntities(
     loadConnectionEntities(c, contactEnts, incPeopleEnts)
   );
 }
+
+export function updateConnectionCache(
```
cainlevy, line 44: this cache maintenance has a very specific effect, yeah?

```suggestion
export function removePendingConnectionRequest
```

cainlevy (approved):

PR #225, opened 2022-01-27
when user selects a voice call to listen in, they connect to the call

gabrielsaruhashi:
Renamed some files in the process! The prettier import sorting made this quite noisy.

Even though I was using node env, I could not prevent the package-lock from flipping to version 2 (hence the number of lines changes 🧐 )

@cainlevy do you know why this might have happened?

## Up Next
DOC still cannot terminate the call. Need to add a nullable binding to CCC


package.json
```diff
@@ -18,13 +18,14 @@
...
     "chart.js": "^2.9.3",
+    "connect-call-client": "^0.2.0",
     "connected-react-router": "^6.8.0",
     "date-fns": "^2.26.0",
     "graphql": "^16.0.0",
     "http-proxy-middleware": "^1.1.1",
     "lodash": "^4.17.20",
     "mediasoup-client": "^3.6.23",
-    "react": "^16.13.1",
+    "react": "^17.0.2",
```
gabrielsaruhashi, line 28: after bumping connect-call-client to 0.2.0, we were forced to bump the react version!

cainlevy: oh no let's fix that with a 0.2.1

cainlevy: i don't think we should bump a dependency this major as part of another pull request. should we make this upgrade separately, or relax connect-call-client's requirements to allow 16.13?

package.json
```diff
@@ -92,7 +93,7 @@
     "@types/jest": "^24.0.0",
     "@types/lodash": "^4.14.165",
     "@types/node": "^12.0.0",
-    "@types/react": "^16.9.0",
+    "@types/react": "^16.14.21",
```
cainlevy, line 96: we want this to match at least the major.minor version of our react dependency

cainlevy:
> Even though I was using node env, I could not prevent the package-lock from flipping to version 2 (hence the number of lines changes 🧐 )

what's the output of running the following in the connect-doc-client directory:

```
nodenv version
npm --version
```

i have nodejs 14.17.4 and npm 6.14.14.

src/components/VoiceCall/VoiceCallAudio.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }, []);
+
+  const handlePeerDisconnected = useCallback((peer: Participant) => {
+    if (peer.type === "doc") return;
+    showToast(
+      "participantConnect",
+      `${capitalize(peer.type)} disconnected.`,
+      "info"
+    );
+  }, []);
```
cainlevy, line 34: are these toasts interesting to doc staff?

src/components/VoiceCall/VoiceCallAudio.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }, [error]);
+
+  useEffect(() => {
+    if (!status) return;
+    showToast(
+      "participantConnect",
+      `Voice call status update: ${status} `,
+      "info"
+    );
+  }, [status]);
```
cainlevy, line 60: i can't think of any status changes that are interesting to users. maybe when the call ends? is that what this toast does?

src/graphql/GetFacilityCalls.graphql
```diff
@@ -0,0 +1,43 @@
...
+            group {
+              id
+            }
+            meetingType
+          }
+          call {
+            ...CallFragment
+            token
+            url
+          }
```
cainlevy, line 38: i'm suddenly wondering how this will impact Kubernetes-as-a-database. here's hoping that our loader strategy pays off!

cainlevy (approved):
we sure make a lot of toast! 🍞 🔥

src/components/VoiceCall/VoiceCallAudio.tsx
gabrielsaruhashi: oh this def won't be on the final version! just a placeholder while we wait for the player
it's pretty helpful for dev lol

src/components/VoiceCall/VoiceCallAudio.tsx
gabrielsaruhashi: naw! they will be replaced by bettter UX when we have the proper player
It's just hard rn to visualize what's going on bc it's audio only during dev lol, so this helps

PR #227, opened 2022-01-28
useFetchAll exhaustively retrieves all pages from a connection

cainlevy:
Example:

```js
  const { data, fetchMore } = useQuery(GetThingsDocument, {
    variables: { myId: my.id }
  });
  useFetchAll(data?.my.things.pageInfo, fetchMore);
```

This depends on the server schema passing through standard pagination args, and on a client merge configuration.

Depends on https://github.com/Ameelio/connect-api-nest/pull/677 for correct operation.


src/utils/useFetchAll.ts
```diff
@@ -0,0 +1,36 @@
...
+export default function useFetchAll<
+  D extends { pageInfo: ForwardPageInfo },
+  V extends { cursor: string }
+>(
+  pageInfo: ForwardPageInfo | undefined,
+  fetchMore: ObservableQueryFields<D, V>["fetchMore"]
+) {
+  useEffect(() => {
+    if (!pageInfo?.hasNextPage) return;
+    fetchMore({
```
lspdrz, line 32: This is interesting. Is this for the inmates gql query I was working on? And will fetchMore trigger N requests?

lspdrz (approved):

src/utils/useFetchAll.ts
lspdrz: Maybe it's ok that multiple requests are made as long as the initial request is fast and the data is almost immediately present, as opposed to waiting for the request for all data to finish. Is that your reasoning here?

cainlevy: I figure this lets us get array-like behavior with a future-proof paginated endpoint. It's a small step from here to full pagination.

PR #228, opened 2022-01-28
refactor with useSelectedFacility

cainlevy:
I wanted this while working on something and discovered @gabrielsaruhashi had already made it!

There are still a few places that read `state.facilities.selected` directly, either because they're interested in the optionality (App) or because they can't use hooks (redux).


lspdrz (approved):
Looks good to me!!!

PR #229, opened 2022-01-28
server-side meeting filters

cainlevy:
The SearchMeetings component now implements all filtering server-side. It only supports filtering on `status` and `scheduledStart`.

image

This is preparation for pagination, but it still improves on the functionality of the page by using date ranges to stay within the `limit: 500`.


src/components/SearchMeetings/SearchMeetings.tsx
```diff
@@ -1,182 +1,92 @@
...
+    value: MeetingStatus.Scheduled,
+  },
+];
+
+const SearchMeetings = ({ facility, meetingType }: Props) => {
+  const [filters, setFilters] = useState<Filters>({
+    meetingStatus: [MeetingStatus.Scheduled],
+  });
+
+  const { loading, error, data, refetch } = useQuery(
```
gabrielsaruhashi, line 52: should we use the reportError util instead of throwing? 
It feels like a better experience than showing that fallback error boundary page

gabrielsaruhashi (approved):
🙌

PR #230, opened 2022-01-28
limit the meetings search results

cainlevy:
This displays an alert when the search filters are too broad to show all of the results. I opted to place it at the top instead of the bottom to reduce the chance of staff being surprised if they generate an incomplete CSV without scrolling all the way down.

image

My hope is that when DOC staff run into this limit, they will communicate use cases that we can solve more directly.


src/components/SearchMeetings/SearchMeetingsTable.tsx
```diff
@@ -57,7 +57,7 @@ const SearchMeetingsTable: React.FC<Props> = ({ meetings, isLoading }) => {
     <Table
       dataSource={meetings}
       loading={isLoading}
-      pagination={{ position: ["topLeft", "bottomLeft"] }}
+      pagination={false}
```
cainlevy, line 60: I disabled pagination because a finite list is more functional. Savvy users can find-in-page if they're hunting for a specific name, without needing a custom interface.

gabrielsaruhashi (commented):
have some concerns about the 100 limit! could we chat on Mon?

cainlevy:
We discussed something like:

- limit the in-page results to 100, so that the page loads in a reasonable period of time
- fetch all results for generating the CSV

gabrielsaruhashi:
@cainlevy let me know when I should look at this again!

src/components/SearchMeetings/SearchMeetings.tsx
```diff
@@ -126,6 +120,12 @@ const SearchMeetings = ({ facility, meetingType }: Props) => {
       />
       <div style={WRAPPER_STYLE}>
         <SearchMeetingsTable meetings={meetings} isLoading={loading} />
+        {data?.facility.meetings.pageInfo.hasNextPage && (
+          <Alert
+            type="warning"
+            message={`Only the first ${LIMIT} results are shown. Please refine your search or export to view all results.`}
+          />
+        )}
```
cainlevy, line 128: i moved this alert back to the bottom of the list where it feels natural. previously i'd placed it at the top of the list because the limit affected exports, but that's no longer the case.

cainlevy (commented):
@gabrielsaruhashi ready!

gabrielsaruhashi (approved):

PR #231, opened 2022-01-31
refactor out of redux modals pattern

cainlevy:
Three places still relied on redux for modals:

1. Reject a connection request
2. Reject a pending meeting
3. Cancel a scheduled meeting

Now they don't.


wittejm (approved):
I struggled with the idea of using "Cancel" for the name for a modal that covers both "Cancel" and "Reject" use cases, but I can't think of a better idea so :shrug:

cainlevy:
Haha yeah I tried to rename the modal partway through my refactor before realizing how it got used and failing to think of a better idea.

PR #232, opened 2022-01-31
convert downloaded reports to graphql native types

cainlevy:
Some of our old Redux types were in my way while working on a new data loading pattern for the downloaded reports. This changeset uses `Pick` to construct a custom type from GraphQL codegen, and then converts from Redux as needed.

Yes, this means that `SearchMeetings.tsx` will query GraphQL, convert to Redux, and then _convert back to GraphQL for some children_. It's the cleanest way I could find to get this refactor out in advance of my upcoming functionality.


src/components/Dashboard/DailyReport/DailyReportMeeting.ts
```diff
@@ -0,0 +1,13 @@
...
+
+type DailyReportMeeting = Pick<Meeting, "id"> & {
+  interval: Pick<Interval, "startAt" | "endAt">;
+  kiosk?: Pick<Kiosk, "name"> | null;
+  visitors: Pick<Visitor, "id" | "firstName" | "lastName">[];
+  inmates: Pick<
+    Inmate,
+    "id" | "inmateIdentification" | "firstName" | "lastName"
+  >[];
+};
```
cainlevy, line 11: things i noticed during my first serious attempt at `Pick`-ing a local type:

1. `or null` on optional fields gets me
2. it's not as plain and readable as redundant types
3. it still feels better than creating a fully isolated / redundant type

src/utils/Call.ts
```diff
@@ -75,23 +75,33 @@ export const meetingsToday = <TCall extends BaseCall>(
...
 
+export type CSVMeeting = Pick<
+  gql.Meeting,
+  "id" | "statusDetails" | "meetingType" | "minors"
+> & {
+  kiosk?: Pick<gql.Kiosk, "name"> | null;
+  interval: Pick<gql.Interval, "startAt" | "endAt">;
+  visitors: Pick<gql.Visitor, "id" | "firstName" | "lastName">[];
+  inmates: Pick<gql.Inmate, "id" | "firstName" | "lastName">[];
+};
```
cainlevy, line 86: this is decoupled from the earlier DailyReportMeeting type. i don't think that's right, but it's a pragmatic step without going deeper into re-organization.

wittejm (approved):
nitpick, but the alternate capitizations of `ql`, `Ql`, and `QL` in various var names and function names is jarring. My mild preference is for `ql` in all cases, but anything consistent and/or intuitive would be nice.

cainlevy:
yeah i can see how that might be a speed bump while reading. my instinct is `QL` because that's part of the official name, but i'm also capable of not caring.

have you heard the one about how to capitalize `id`? 😂

cainlevy:
just to explore the depths of my personal inconsistency:

- i choose `gql` as an object because i reserve titlecase for constants, classes, and types. `gQL` offends me.
- i choose `graphQL` when i'm in the mood for a longer const name
- i choose `graphql` in normal async conversation unless i'm caffeinated enough to deploy my shift key reliably, and then choose `GraphQL`

so i guess i'm just going to veto `gQl` 🙈

PR #235, opened 2022-02-01
fix PDF generation by fixing Date scalar's type

cainlevy:
Previously the `Date` scalar from GraphQL was typed as `any`. That became a problem when we started seeing `Invalid time value` errors caused by attempting to use a number as a string.

This is a thorough update. I'll flag the offending line.


src/components/Dashboard/DailyReport/DailyReport.tsx
```diff
@@ -33,17 +32,13 @@ const DailyReport: React.FC<Props> = ({
...
-    Record<string, DailyReportMeeting[]>
-  >({});
-
-  useEffect(() => {
-    const blocks: Record<string, DailyReportMeeting[]> = groupBy(
-      meetings,
-      (call) => call.interval.startAt
-    );
-    setCallBlocks(blocks);
-  }, [meetings]);
```
cainlevy, line 46: this was the source of our bug. `interval.startAt` is a `number` but gets typecast to `string` through some presumptive`groupBy` generics. later, we do something that requires it to be a string and error.

wittejm (approved):

PR #236, opened 2022-02-02
replace download menu with export modal

cainlevy:
https://user-images.githubusercontent.com/4449/152214621-28f7edb1-e937-4f65-aa06-f69e1d6fe695.mov


src/api/client.ts
```diff
@@ -18,7 +19,22 @@ const authLink = setContext((_, { headers }) => {
 
 export const client = new ApolloClient({
   link: authLink.concat(httpLink),
-  cache: new InMemoryCache(),
+  cache: new InMemoryCache({
+    typePolicies: {
+      Meeting: {
+        fields: {
+          interval: {
+            merge: true,
```
gabrielsaruhashi, line 27: oh what does this merge policy do?

src/utils/useFetchAll.ts
```diff
@@ -23,12 +23,20 @@ type ForwardPageInfo = Pick<PageInfo, "endCursor" | "hasNextPage">;
...
+  pageInfo: ForwardPageInfo | undefined;
+  fetchMore: ObservableQueryFields<D, V>["fetchMore"];
+  onComplete?: () => void;
+}) {
   useEffect(() => {
-    if (!pageInfo?.hasNextPage) return;
+    if (!pageInfo?.hasNextPage) {
+      if (onComplete) onComplete();
+      return;
+    }
```
gabrielsaruhashi, line 39: oh wow this is cool. so it abstracts all the new paginated fetching 🤯

src/components/ExportSchedule/ExportScheduleModal.tsx
```diff
@@ -0,0 +1,148 @@
...
+      meetingType: [meetingType],
+    },
+  });
+
+  const saveAndClose = () => {
+    setGenerating(true);
+    if (format === Format.PDF) {
+      // give react time to update the button before locking up the page
+      setTimeout(() => {
+        pdf(
```
gabrielsaruhashi, line 67: this is being nixed on the next PR I imagine

src/components/ExportSchedule/useFullSchedule.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const { data, fetchMore } = useQuery(GetScheduleMeetingsDocument, {
+    variables: {
+      id: facilityId,
+      ...filters,
+    },
+    onError: errorReporter("useFullSchedule"),
+  });
+
+  useFetchAll({
```
gabrielsaruhashi, line 31: I see - so this hook automatically handles the pagination request and data is updated with the new page's info

src/components/ExportSchedule/useFullSchedule.ts
```diff
@@ -0,0 +1,43 @@
...
+    pageInfo: data?.facility.meetings.pageInfo,
+    fetchMore,
+    onComplete: () => setLoading(false),
+  });
+
+  return {
+    loading,
+    data: loading ? [] : data?.facility.meetings.edges.map((e) => e.node) || [],
```
gabrielsaruhashi, line 39: does this mean that every time we fetch a new page, the array will appear fully empty for a second? before re-appearing with the new page

gabrielsaruhashi (approved):
alt

src/api/client.ts
cainlevy: `merge: true` tells apollo to merge different values even if it can't verify that the object has the same identity. it can be useful for graphql types that don't come from the database.

in my case this merges `{startAt, endAt}` from my query with `{startAt, endAt, seconds}` from another query. i could have also solved the problem by adding "seconds" to my result, but i was in the mood to query the minimum data necessary.

src/utils/useFetchAll.ts
cainlevy: yep! this lets us escape pagination without creating api debt.

src/components/ExportSchedule/useFullSchedule.ts
cainlevy: this means that we won't return any data until we have _all_ of the pages.

the pages will appear to be a single array because of how merging is configured for this field in our cache.

PR #238, opened 2022-02-02
archive PDF schedule

cainlevy:
The current PDF report has a high cost. It does not scale to even a single day's meetings at a medium facility, requires a [large library](https://bundlephobia.com/package/@react-pdf/renderer@2.1.1), and provides little in the way usability over the CSV alternative.

image

This changeset archives all of the PDF schedule functionality. If we revisit, it will be right here in our git history.


gabrielsaruhashi (approved):
was fun to work on this back in the days ~sad reacts

PR #241, opened 2022-02-09
reimplement against the published REST endpoint

cainlevy:
Meant to include this in the previous pull request, but it probably deserves its own peek anyway.


src/components/ImportRoster/ValidatedResult.tsx
```diff
@@ -1,44 +1,38 @@
...
+  const [submitting, setSubmitting] = useState(false);
+
   const handleSubmit = async () => {
-    await upsertInmates({
-      variables: {
-        input: {
-          facilityId: facility.id,
+    setSubmitting(true);
+    try {
+      await fetchAuthenticated("uploadMembers", {
```
cainlevy, line 20: i think this could go a long ways towards assuring people that automation uses the same process

gabrielsaruhashi: oh I see - I imagine it must be odd to give a graphql spec for the DOC to hit by themselves

gabrielsaruhashi (approved):

PR #244, opened 2022-02-11
add "go back" to error boundary so we can use it on pages

cainlevy:
I've improved the UX of our error boundary with a "go back" button that should be able to recover from bad page loads.

This is useful for pages that do not have an "empty" state and therefore can't get away with simply toasting a message with `onError`. Otherwise, those pages can get trapped with an infinite loader.

image


src/pages/InmatePage.tsx
```diff
@@ -51,9 +49,11 @@ function InmatePage({ match }: RouteComponentProps<TParams>): ReactElement {
     dispatch(fetchInPersonVisits({ "inmateParticipants.inmateId": inmateId }));
   }, [inmateId, dispatch]);
 
-  const connections = data?.inmate.connections;
+  if (error) throw error;
+  if (loading || !data) return <PageLoader />;
 
-  if (!inmate || !connections) return <PageLoader />;
```
cainlevy, line 56: this is where the page got stuck with an infinite spinner. it couldn't distinguish between "no data because loading" and "no data because error".

gabrielsaruhashi (approved):

PR #250, opened 2022-02-23
admins may import rosters

cainlevy:
Release the import modal to facility admins. 😰


src/pages/InmatesPage.tsx
```diff
@@ -103,7 +108,7 @@ const InmatesPage: React.FC = () => {
               onChange={(event) => setSearchQuery(event.target.value)}
             />
           </Input.Group>,
-          isSuperUser(session.user) ? (
+          role === "admin" ? (
```
wittejm, line 111: Looks like everything except this line is a refactor. Is that correct?

wittejm (approved):

src/pages/InmatesPage.tsx
cainlevy: yes that's right.

would it have been helpful in this case to split up the pull request or comment on what happened?

wittejm: Nope it's fine! It let me do a little bit more of my own code comprehension work.

PR #251, opened 2022-02-24
Conflicting DoB graphql fix

wittejm:
This leaves DOB in the fragments, and instead uses custom inline fragments on `sender` field which is a `VisitorOrInmate` schema type.


wittejm:
Tested locally: with this fix, the Pending Messages and Connection Messages views both load for me now.

gabrielsaruhashi (approved):

PR #257, opened 2022-03-03
Inactive keywords behavior

wittejm:
- Button to show/hide inactive keywords in list
- Inactive keywords are greyed and marked "(INACTIVE)"
- replace the "delete" button with "deactivate"/"reactivate" buttons.

Really makes me wonder if I should have made this with one "toggle" mutation. Sigh.


wittejm:
forgot to add the vid... uploaded now

src/components/Settings/KeywordManagementSettings.tsx
```diff
@@ -152,11 +190,19 @@ const KeywordManagementSettings = (props: Props) => {
...
+                  >
+                    Deactivate
+                  </Typography.Link>
+                ) : (
+                  <Typography.Link
+                    onClick={() => setReactivatingKeyword(keyword)}
+                  >
+                    Reactivate
+                  </Typography.Link>
+                )}
```
cainlevy, line 205: i think having separate mutations is nice on the backend, because of the high potential for differences in side effects. but if you like the feeling of a toggle, you could wrap these two links into a single component with private mutations and modals.

cainlevy (approved):

src/components/Settings/KeywordManagementSettings.tsx
wittejm: Gotcha. I just messed with this idea for a bit but it creates some extra conditionals that still looks gross, so, it's probably not worth the effort.

PR #258, opened 2022-03-03
bugfix for keyword self-collision

wittejm:
I introduced this bug right at the end of my work after doing testing. 

It was causing an error with editing an existing keyword's alternate spellings because it made this filter fail: 

https://github.com/Ameelio/connect-doc-client/blob/3302c2567c5a7d6f3249ba1f42af68d7a14109bc/src/components/Settings/AddEditKeywordModal.tsx#L183

very fun bug hunt.


src/components/Settings/KeywordManagementSettings.tsx
```diff
@@ -113,7 +113,7 @@ const KeywordManagementSettings = (props: Props) => {
         dataSource={filteredKeywords.map((keyword) => ({
           key: keyword.id,
           ...keyword,
-          name: `${keyword.name} ${keyword.isActive ? "" : " (INACTIVE)"}`,
```
cainlevy, line 116: i'm not entirely sure how this data set finds its way to the modal, but removing an extra space char seems like a fine thing

cainlevy (approved):

src/components/Settings/KeywordManagementSettings.tsx
wittejm: Yeah, it makes the whole thing feel a little hacky. But now everything works!

PR #262, opened 2022-03-04
redux dead code

cainlevy:
I found a number of selectors and thunks and the like that are not currently used. Less to migrate!


wittejm (approved):
Reviewing a PR like this I always just murmur to myself, "mmmm free stuff, free stuff..."

PR #275, opened 2022-03-15
rename time quota field

wittejm:
https://github.com/Ameelio/connect-api-nest/pull/797

Does this change need to be synchronous? I'm probably being unnecessarily cautious here; nothing breaks due to a field missing from the graphql server if it's not used for anything, right? 

...of course not. But I've come this far.


wittejm:
hm? This is just the client. the schema.graphql change is in the linked api PR

PR #280, opened 2022-03-29
fetch only facility's connections on visitor

wittejm:
Fix forbidden errors for when a visitor has connections from different facility systems.


wittejm:
I didn't test this one with custom data :(  but the fix seems simple enough, and at least the page loads correctly :) .

cainlevy (commented):
i think we should additionally filter the visitor's meetings to only show ones from the current facility.

it could get a little odd since connections "transfer" (mutable `facilityId`) while meetings do not (immutable `facilityId`), but the page will just have to make do.

wittejm:
Added this to meeting as well. Graphql confuses me as to what it generates where: I added this second instance of facilityId in the query and `npm run graphql` only updated graphql.ts, not the generated query document.

cainlevy (approved):

PR #283, opened 2022-04-07
remove unused redux thunks

cainlevy:
Next up: deleting unused NestJS code so I don't update the wrong function again.


wittejm (approved):
Woohoo! Burn, redux, burn!

PR #284, opened 2022-04-07
disable connection uses  rejection modal

gabrielsaruhashi:
image


gabrielsaruhashi:
@cshdllr 
took the liberty to also update the connection drawer with this flow

cshdllr:
Content nits...
1. Can we please put "Not an approved visitor" at the top of the list since it is overwhelmingly the #1 rejection reason?
2. Can we use sentence case for the options? (like below)

• Not an approved visitor
• Photo of ID is blurry or cut off
• Incomplete information
• Other

src/components/Drawer/ConnectionDetailDrawer.tsx
```diff
@@ -62,22 +78,34 @@ const ApproveRejectButton: React.FC<ApproveRejectButtonProps> = ({
...
+                      statusDetails: feedback,
+                    },
+                  },
+                })
+              }
+              onClose={() => setIsRejecting(false)}
+              connection={connection}
+              facility={facility}
+            />
+          ) : (
```
cainlevy, line 99: i think typically we would continue rendering the button

cainlevy (approved):

src/components/Drawer/ConnectionDetailDrawer.tsx
gabrielsaruhashi: oop, was an oversight

PR #288, opened 2022-04-25
create kiosk with multiple groups

cainlevy:
Depends on https://github.com/Ameelio/connect-api-nest/pull/868

Switching to GraphQL allows us to select exactly which groups are associated with a kiosk. We must also fetch kiosks from GraphQL to see the new data structure ("Location" becomes "Groups" in the table).

https://user-images.githubusercontent.com/4449/165177329-a6fe234c-f955-4620-b3e8-d917945873a1.mov


src/api/client.ts
```diff
@@ -66,3 +67,14 @@ export const urqlClient = createClient({
   },
   exchanges: [fetchExchange],
 });
+
+/**
+ * appendItem will return a cache modifier that adds an object to the end
+ * of a list.
+ */
+export function appendItem<T extends { id: string }>(object: T): Modifier<T[]> {
```
wittejm, line 75: ooh nice

wittejm (approved):

PR #291, opened 2022-04-28
fixed faulty get visitor query

gabrielsaruhashi:
image


src/graphql/GetVisitor.graphql
```diff
@@ -1,7 +1,7 @@
 query GetVisitor($visitorId: ID!, $facilityId: ID!) {
   visitor(id: $visitorId) {
     ...VisitorFragment
-    connections(facilityId : $facilityId}) {
+    connections(facilityId: $facilityId) {
```
gabrielsaruhashi, line 4: i'm surprised there was a typo here in prod
was driving me crazy trying to figure out why it was not updating the .generated ts query

src/graphql/GetVisitor.graphql
```diff
@@ -50,7 +50,7 @@ query GetVisitor($visitorId: ID!, $facilityId: ID!) {
             name
             description
             enabled
-            group {
+            groups {
```
wittejm, line 53: Groups has an id? Isn't it an array?

wittejm (approved):

src/graphql/GetVisitor.graphql
gabrielsaruhashi: hm? this is the syntax for it regardless of whether or not is an array. what matters is this ent
https://graphql.org/learn/queries/

crcommons (approved):

src/graphql/GetVisitor.graphql
cainlevy: this must be why i didn't discover the bug before changing the schema 😭

src/graphql/GetVisitor.graphql
cainlevy: we can remove `groups { id }`. it's unused.

PR #293, opened 2022-04-28
organize imports on commit

cainlevy:
see https://github.com/Ameelio/connect-api-nest/pull/874


wittejm (approved):
Just adding the package dep creates a commit hook?

cainlevy:
We already have a prettier commit hook, so this makes it better 🎉 .

Oh, actually, DOC client's husky version is pretty far out of date. That might explain some things. Ugh. So many repositories.

cainlevy:
updated husky and lint-staged as well.

PR #299, opened 2022-05-05
schedule displays phone and number

gabrielsaruhashi:


brandon-bee (approved):
Thanks for making this change so quickly!

PR #300, opened 2022-05-05
scroll to focus on every render

wittejm:
The current behavior is that the scroll into focus doesn't happen on pageload. That's because the useEffect is called before the first render, before the focusRef has been set for the first time. This change reruns the focus after the message has rendered. 

This feels like the wrong solution. I really want to run the useEffect only one extra time: upon first page load after the first render. 

Since it now runs on every render, it refreshes focus again to the top tagged message when the page polls for new messages (every 15 seconds). This is potentially annoying.


src/components/MessageReviewPanel/MessagesList.tsx
```diff
@@ -18,7 +18,7 @@ const MessagesList = ({ messages, focus }: Props) => {
       behavior: "smooth",
       block: "start",
     });
-  }, [focusRef, focus?.id]);
+  });
```
cainlevy, line 21: `[focusRef.current]` might prevent this from scrolling too frequently, but since we're relying on a ref instead of state there's no guarantee that the function will re-render once our focusRef exists.

wittejm: I did try focusRef.current and that didn't work, Typescript said this:  `Mutable values like 'focusRef.current' aren't valid dependencies because mutating them doesn't re-render the component.`

cainlevy: hah, yes, quite right.

however, removing them from the dependency array doesn't re-render the component either. something else may currently take care of it by accident, but i expect it's fragile.

wittejm: Do you have thoughts on a better solution? Liz said if this behavior wasn't possible, it would be ok to make the "Approve" button initially say "Review" and take you to the first focused message.

cainlevy: i haven't forgotten this! i don't know an immediate solution, but the behavior you described in the PR where the window scrolls every 15 seconds doesn't sound great either.

wittejm: Yep. I'll fiddle around with it a bit (it seems like it just takes some smarter use of useEffect) and if I don't make headway I'll  just do the second option Liz suggested.

wittejm:
@cainlevy Found the correct approach, pointed to here 
https://stackoverflow.com/questions/60476155/is-it-safe-to-use-ref-current-as-useeffects-dependency-when-ref-points-to-a-dom
and explained further here: 
https://medium.com/welldone-software/usecallback-might-be-what-you-meant-by-useref-useeffect-773bc0278ae

This intermediate commit is a workaround using setTimeout that I felt pretty proud of for a hot minute. But I'll just squash this on merge.

src/components/MessageReviewPanel/MessagesList.tsx
```diff
@@ -10,15 +10,12 @@ interface Props {
...
-  useEffect(() => {
-    if (!focusRef.current) return;
-    focusRef.current.scrollIntoView({
+  const scrollToFocus = useCallback((node) => {
+    node.scrollIntoView({
       behavior: "smooth",
       block: "start",
     });
-  }, [focusRef, focus?.id]);
+  }, []);
```
cainlevy, line 18: wow this is great!

cainlevy (approved):
amazing find. i'll bet we can use this pattern more.

PR #303, opened 2022-05-11
schedule export: email and phone are only available to super admins

gabrielsaruhashi:


src/components/ExportSchedule/ExportScheduleModal.tsx
```diff
@@ -86,11 +99,14 @@ const ExportScheduleModal: React.FC<Props> = ({
...
+              onChange={(e) => setReportType(e.target.value)}
+              value={reportType}
             >
-              <Radio value={false}>Public</Radio>
-              <Radio value={true}>Admin</Radio>
+              <RadioReportType value="public" label="Public" />
+              <RadioReportType value="admin" label="Admin" />
+              {isSuperUser(session.user) && (
+                <RadioReportType value="superadmin" label="Ameelio Admin" />
+              )}
```
cainlevy, line 109: what's the use case for ameelio needing its own export type?

cainlevy (approved):

src/components/ExportSchedule/ExportScheduleModal.tsx
gabrielsaruhashi: @brandon-bee needs to notify users from time to time about some event / update that is coming up

PR #305, opened 2022-05-12
Expired Token Error Handling

crcommons:
- adds error link for error checking with graphql and apollo so that we can read error messages from api
- checks for expired token errors for rest endpoints, displays toast, and logs user out
- checks for expired tokens for graphql requests, displays toast, and logs user out


cainlevy (approved):
🙌

cainlevy:
changelog?

crcommons:
It's not actually live because the API isn't merged yet

PR #307, opened 2022-05-16
fix and migrate meeting detail page

cainlevy:
Depends on https://github.com/Ameelio/connect-api-nest/pull/905

The meeting detail page (aka RecordingPage?) is available via "View Details" while looking at the meeting table for one of the participants. It has been broken for quite some time. Now it looks like this:

image


src/pages/RecordingPage.tsx
```diff
@@ -45,60 +27,62 @@ function RecordingBase({
...
+        {/* {call.recordingPath && (
           <ReactPlayer
             autoplay={true}
             muted={true}
             controls={true}
             width="100%"
             url={call.recordingPath}
           />
-        )}
+        )} */}
```
cainlevy, line 48: another day

src/pages/RecordingPage.tsx
```diff
@@ -45,60 +27,62 @@ function RecordingBase({
...
         <PageHeader
           ghost={false}
           onBack={() => window.history.back()}
-          title={`Call ${data?.meeting.call.id}`}
-          subTitle={facility}
```
cainlevy, line 71: react was pretty upset about being asked to render a plain object

src/pages/RecordingPage.tsx
```diff
@@ -45,60 +27,62 @@ function RecordingBase({
...
+          title={`Call ${meeting.id}`}
+          subTitle={facility.fullName}
         >
-          <Descriptions size="small" column={3} bordered layout="vertical">
+          <Descriptions size="small" column={1} bordered>
             <Descriptions.Item label="Visitation ID">
-              {call.id}
+              {meeting.id}
```
wittejm, line 57: are uuids useful to show in the client? Is there a better way? We do this elsewhere too and it just seems like horrible UX.

src/utils/graphqlConversions.ts
```diff
@@ -165,15 +165,23 @@ export const graphQlMeetingToVisit = (meeting: GraphqlMeeting): Visit => {
 };
 
 export const graphqlCallMessageToCallMessage = (
```
wittejm, line 167: conversions were only needed because of redux right? Is this slated for removal too?

src/utils/graphqlConversions.ts
```diff
@@ -165,15 +165,23 @@ export const graphQlMeetingToVisit = (meeting: GraphqlMeeting): Visit => {
...
-    senderType: message.sender.__typename === "Inmate" ? "inmate" : "user",
+    senderType:
+      message.sender.__typename === "Inmate"
+        ? "inmate"
+        : message.sender.__typename === "Visitor"
+        ? "user"
+        : "doc",
     contents: message.contents,
-    createdAt: message.createdAt,
+    createdAt: new Date(message.createdAt).getTime(),
```
wittejm, line 184: What's this doing? Isn't `createdAt` already a numeric timestamp?

wittejm (approved):

src/pages/RecordingPage.tsx
cainlevy: We currently ask people to reference UUIDs when reporting issues to us or asking for recordings. I agree it's far from ideal.

On the up side, UUIDs can make things look Very Official and like Serious Business because of how intimidatingly opaque they are.

src/utils/graphqlConversions.ts
cainlevy: that's what i thought! then i discovered in this case we're actually converting to the CVH format of a CallMessage. we may be stuck with this logic in some form, even if it doesn't live here.

src/utils/graphqlConversions.ts
cainlevy: I think ... this is a leftover from when I was playing with a string type. Will simplify.

PR #308, opened 2022-05-17
rely on GraphQL for contact page

cainlevy:
The biggest consideration here is the TTL on visitor identification images. The URLs are valid for 60 minutes, which should be plenty for normal usage. If we get complaints we will want to switch to JIT loading.


cainlevy:
oops this got the dashboard code. needs to go before a meeting today.

wittejm (approved):

PR #310, opened 2022-05-19
stop parsing dateOfBirth as UTC

cainlevy:
> When the time zone offset is absent, date-only forms are interpreted as a UTC time and date-time forms are interpreted as local time.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse


cainlevy:
@wittejm this bug is causing visitors to be rejected

PR #313, opened 2022-05-26
upgrade to ccc@0.9.0

cainlevy:
Upgrades voice call monitoring to use the latest connect-call-client.


src/components/VoiceCall/VoiceCallPlayer.tsx
```diff
@@ -42,22 +39,15 @@ const VoiceCallPlayer = ({
...
-          {!isCallEmpty ? (
+          <InitialsAvatarFacepile people={participants} />
+          {participants.length > 0 ? (
             <Typography.Text type="secondary">
-              {inmateOn ? inmates : ""}
-              {allParticipantsOn ? " & " : ""}
-              {visitorOn ? visitors : ""}
-              {allParticipantsOn ? " are both " : " is "}
+              {participants.map((p) => getFullNames([p])).join(" & ")}
+              {participants.length > 1 ? " are both " : " is "}
```
jackofsometrade, line 50: should we just say `are` instead of `are both`?  IDK how soon we'll have 3+ people, but it fits `> 1` more universally I think.

jackofsometrade (approved):

src/components/VoiceCall/VoiceCallPlayer.tsx
cainlevy: yeah. this was a "not worse!" moment for me. pluralization is so annoying and i don't know why we need to construct an english sentence at all.

PR #314, opened 2022-05-26
use connect-call-client for video!

cainlevy:
It's here! We can use connect-call-client for DOC video monitoring through a clever React trick that I like to call "cheating". It all happens in `CallRegistry` below.

image


src/components/LiveCall/CallRegistry.tsx
```diff
@@ -0,0 +1,87 @@
...
+    []
+  );
+
+  return (
+    <CallRegistryContext.Provider value={registry}>
+      {calls.map((c) => (
+        <RegisterCall key={c.id} call={c} register={register} drop={drop} />
+      ))}
+      {children}
+    </CallRegistryContext.Provider>
```
cainlevy, line 49: welcome to CallRegistry, where we rely on a child component to run a hook for us and then store the results in context!

we have to do this because:

1. the DOC client must remain connected to calls while people multi-task their way around other pages. apparently that counts as "monitoring" even though you can't see or hear.
2. React does not support calling a hook inside of a loop, but it does support calling a hook inside of a component inside of a loop.

src/pages/LiveCallPage/VideoChat.tsx
```diff
@@ -139,31 +92,19 @@ const VideoChat: React.FC<Props> = React.memo(
...
-
-          const isVideoPaused = isInmate
-            ? inmateVideoPaused
-            : contactVideoPaused;
-
-          const name = isInmate
-            ? participantNames.inmates
-            : participantNames.contacts;
+        {connectCall.peers.map((peer, index) => {
+          const name = participantNamesById[peer.user.id];
```
cainlevy, line 96: i might have fixed something here but i didn't test multiple visitors on a call to find out. in theory this will show the correct name for each visitor, instead of showing all visitor names on each visitor.

cainlevy:
I nearly forgot that this depends on a new release of CVH to fix an issue with heartbeat responses.

src/components/LiveCall/CallRegistry.tsx
jackofsometrade: Sounds like we could use a popout component that can float over the rest of the app while the staff member multitasks.

cainlevy: definitely!

in that case we may still need the registry so video streams can transition between the popout and the page. not sure. maybe there's another clever solution.

jackofsometrade (approved):

cainlevy:
I ran through a quick call locally to make sure nothing obvious had broken. Planning to deploy in a couple hours.

PR #320, opened 2022-06-23
render a review modal for org requests

cainlevy:
The new ReviewModal component needs styling and a final home (it's currently organized with `src/Organizations`).

image


wittejm (approved):
Is that the email address of the org or that admin? The organization has its own email and I think that belongs in its own row. Maybe still include the admin's email as well?

PR #326, opened 2022-07-05
Update `Settings` page to edit `Schedules` instead of meeting slots

dabbler0:
Currently hidden behind the `SCHEDULES` feature flag. This:
 - Replaces the current two "meeting slots" tabs in the Settings page with a single "schedules" tab
 - Allows adding and editing schedules (kiosks, groups, name) and meeting slots in that tab with two new drawers.
 - Disables editing `group`s of `Kiosk`s.


src/components/Settings/AddScheduleDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+          label="Meeting Type"
+          name="meetingType"
+          rules={[{ required: true, message: "Meeting Type is required." }]}
+        >
+          <Radio.Group value={selectedMeetingType}>
+            <Space direction="vertical">
+              {[
+                MeetingType.Call,
+                MeetingType.InPersonContact,
+                MeetingType.InPersonNoContact,
```
crcommons, line 92: Is there a reason you manually constructed this array? If not, let's go with `Object.values(MeetingType)` to get the array, which will future proof this

src/components/Settings/AddScheduleDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+              ].map((meetingType) => {
+                return (
+                  <Radio.Button
+                    key={meetingType}
```
crcommons, line 96: In case there were every two values in meetingType that were the same, consider using the index as the key instead

src/components/Settings/AddScheduleDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+                  </Radio.Button>
+                );
+              })}
+            </Space>
+          </Radio.Group>
+        </Form.Item>
+        <Form.Item label="Groups" name="groupIds">
+          <Select mode="multiple" allowClear placeholder="Select...">
+            {groups.map((group) => (
+              <Select.Option value={group.id}>{group.name}</Select.Option>
```
crcommons, line 115: missing `key` which is probably throwing a warning in the console

src/components/Settings/AddScheduleDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+            ))}
+          </Select>
+        </Form.Item>
+        <Form.Item label="Kiosks" name="kioskIds">
+          <Select mode="multiple" allowClear placeholder="Select...">
+            {kiosks.map((kiosk) => (
+              <Select.Option value={kiosk.id}>{kiosk.name}</Select.Option>
```
crcommons, line 122: same here, missing `key`

src/components/Settings/EditKioskDrawer.tsx
```diff
@@ -98,13 +112,15 @@ const EditKioskForm: React.FC<Props> = ({ kiosk, groups, onClose }) => {
...
-          {groups.map((group) => (
-            <Select.Option value={group.id}>{group.name}</Select.Option>
-          ))}
-        </Select>
-      </Form.Item>
+      {!featureEnabled("SCHEDULES") && (
+        <Form.Item label="Groups" name="groupIds">
+          <Select mode="multiple" allowClear placeholder="Select...">
+            {groups.map((group) => (
+              <Select.Option value={group.id}>{group.name}</Select.Option>
```
crcommons, line 119: `key` missing here as well

src/components/Settings/EditScheduleDrawer.tsx
```diff
@@ -0,0 +1,139 @@
...
+        </Select>
+      </Form.Item>
+      <Form.Item label="Kiosks" name="kioskIds">
+        <Select mode="multiple" allowClear placeholder="Select...">
+          {kiosks
+            .filter((kiosk) => {
+              return kiosk.meetingType === schedule.meetingType;
+            })
+            .map((kiosk) => (
+              <Select.Option value={kiosk.id}>{kiosk.name}</Select.Option>
```
crcommons, line 98: `key`

src/components/Settings/EditScheduleDrawer.tsx
```diff
@@ -0,0 +1,139 @@
...
+        label="Name"
+        name="name"
+        rules={[{ required: true, message: "Name is required." }]}
+      >
+        <Input placeholder="Name" />
+      </Form.Item>
+      <Form.Item label="Groups" name="groupIds">
+        <Select mode="multiple" allowClear placeholder="Select...">
+          {groups.map((group) => (
+            <Select.Option value={group.id}>{group.name}</Select.Option>
```
crcommons, line 87: `key`

src/components/Settings/KioskSettings.tsx
```diff
@@ -42,7 +48,9 @@ const KioskSettings: React.FC<Props> = () => {
       dataIndex: "name",
       key: "name",
     },
-    {
+  ];
+  if (!featureEnabled("SCHEDULES")) {
+    columns.push({
```
crcommons, line 53: Just for consideration.... for visual clarity and code readability, you can completely construct the array without the `groups` object and then after it is defined use `column.splice(index, 0, item);` swapping out index and item for the values you want. This way you won't need to break up the array definition.

src/components/Settings/MeetingSlotDrawer.tsx
```diff
@@ -0,0 +1,319 @@
...
+  <Drawer
+    visible={visible}
+    onClose={onClose}
+    closable={false}
+    title="Edit Schedule"
+    placement="right"
+    width="40%"
+  >
+    {schedule ? (
+      <ScheduleMeetingSlotSettings onClose={onClose} schedule={schedule} />
```
crcommons, line 315: Given the length of this file, it would be nice to move ScheduleMeetingSlotSettings to another file.

crcommons (changes requested):
Woah, this is a lot of code!! nice job! 

I did my best to review in depth, but not positive I caught everything. If there are any areas where you have concerns, let me know and I will give an extra thorough review. Otherwise, a few small issues, but nothing major. 

My only question is, if all of the changes have been implemented and deployed in the API, do we need the feature flag?

dabbler0:
A couple outstanding PRs in `connect-api-nest` should be merged before we enable the feature flag -- #1004, #1005 (necessary, fixes some parts of the new api that these doc changes depend on), and #1006 (not necessary, but good for quality of life because without it sometimes there is an overflow of new schedules).

crcommons:
> A couple outstanding PRs in `connect-api-nest` should be merged before we enable the feature flag -- #1004, #1005 (necessary, fixes some parts of the new api that these doc changes depend on), and #1006 (not necessary, but good for quality of life because without it sometimes there is an overflow of new schedules).

ok sounds good. just wanted to check since that original PR was merged

src/components/Settings/AddScheduleDrawer.tsx
dabbler0: This was just copied from `AddKioskDrawer`, but I think it is because we need to exclude voice calls.

src/components/Settings/AddScheduleDrawer.tsx
dabbler0: Fixed this in all the `Selects`, but am still getting a "needs unique key" warning in the console. The warning seems to be coming from the `<Table>` in `KioskSettings` and `ScheduleSettings` (if I remove that table but keep the drawers then the error goes away), but I can't figure out what I'm doing wrong.

dabbler0: Ah, what was needed was `rowKey` on the antd `Table`. Should be all resolved now.

src/components/Settings/AddScheduleDrawer.tsx
```diff
@@ -0,0 +1,221 @@
...
+                MeetingType.InPersonContact,
+                MeetingType.InPersonNoContact,
+              ].map((meetingType) => {
+                return (
+                  <Radio.Button
+                    key={meetingType}
+                    onClick={(e) => {
+                      if (meetingType !== selectedMeetingType) {
+                        setSelectedKiosks([]);
+                      }
```
cainlevy, line 166: this kind of logic is sometimes handled with a `useEffect`, e.g.

```
// clear kiosks when the meeting type changes
useEffect(() => {
  setSelectedKiosks([]);
}, [selectedMeetingType]);
```

i don't think this is worth changing now but i wanted to share so you have options next time.

src/components/Settings/CreateSchedule.graphql
```diff
@@ -0,0 +1,24 @@
+mutation CreateSchedule($input: CreateScheduleInput!) {
+  createSchedule(input: $input) {
+    schedule {
+      ...ScheduleFragment
+      meetingType
```
cainlevy, line 5: could move to `ScheduleFragment`

src/components/Settings/CreateSchedule.graphql
```diff
@@ -0,0 +1,24 @@
...
+      meetingSlots {
+        id
+        day
+        hour
+        minute
+        duration
+      }
+      visitorTypes
```
cainlevy, line 13: could move to `ScheduleFragment`

src/components/Settings/EditScheduleDrawer.tsx
```diff
@@ -0,0 +1,440 @@
...
+    "id" | "name" | "meetingType" | "meetingSlots" | "visitorTypes"
+  > & {
+    groups: Pick<Group, "id" | "name">[];
+    kiosks: Pick<Kiosk, "id" | "name">[];
+  };
+  groups: (Pick<Group, "id" | "name"> & { parent?: { id: string } | null })[];
+  kiosks: (Pick<Kiosk, "id" | "name" | "meetingType"> & {
+    parent?: { id: string } | null;
+  })[];
+}
```
cainlevy, line 50: given how much is going on in this file, it may be worth extracting `DetailsProps` and `ScheduleDetailsTable` into a separate file

src/components/Settings/EditScheduleDrawer.tsx
```diff
@@ -0,0 +1,440 @@
...
+    groups: Pick<Group, "id" | "name">[];
+    kiosks: Pick<Kiosk, "id" | "name">[];
+  };
+  groups: (Pick<Group, "id" | "name"> & { parent?: { id: string } | null })[];
+  form: FormInstance;
+  kiosks: (Pick<Kiosk, "id" | "name" | "meetingType"> & {
+    parent?: { id: string } | null;
+  })[];
+  onClose: () => void;
+}
```
cainlevy, line 66: i would also encourage extracting `FormProps` and `EditScheduleForm` into a separate file

src/components/Settings/ScheduleSettings.tsx
```diff
@@ -0,0 +1,182 @@
...
+  type: VisitorType;
+}
+
+const VisitorTypeTag: React.FC<VisitorTypeTagProps> = ({ type }) => {
+  if (type === VisitorType.Professional) {
+    return <Tag>Professional</Tag>;
+  } else {
+    return <Tag>Personal</Tag>;
+  }
+};
```
cainlevy, line 32: this looks like it could be extracted and shared with `EditScheduleDrawer`.

src/components/Settings/ScheduleSettings.tsx
```diff
@@ -0,0 +1,182 @@
...
+                ))}
+              </div>
+            }
+          >
+            <Tag key="etc">+{items.length - 2} others</Tag>
+          </Tooltip>
+        ))}
+    </>
+  );
+};
```
cainlevy, line 61: this seems like a nice generic utility component. we don't have a specific location for utility components in the DOC client so i started adding them to `src/components` but i think `src/utils` may be better.

src/components/Settings/GetFacilitySchedules.graphql
```diff
@@ -0,0 +1,34 @@
+query GetFacilitySchedules($facilityId: ID!) {
+  facility(id: $facilityId) {
+    id
+    schedules {
+      ...ScheduleFragment
+      meetingType
```
cainlevy, line 6: could move to `ScheduleFragment`

src/components/Settings/GetFacilitySchedules.graphql
```diff
@@ -0,0 +1,34 @@
...
+      meetingSlots {
+        id
+        day
+        hour
+        minute
+        duration
+      }
+      visitorTypes
```
cainlevy, line 14: could move to `ScheduleFragment`

src/components/Settings/ScheduleMeetingSlotSettings.tsx
```diff
@@ -0,0 +1,222 @@
...
+} {
+  const start = nextDay(new Date(), slot.day as Day);
+  start.setHours(slot.hour);
+  start.setMinutes(slot.minute);
+  start.setSeconds(0);
+  start.setMilliseconds(0);
+
+  const end = addMilliseconds(start, slot.duration);
+  return { start, end };
+}
```
cainlevy, line 79: i think i saw this earlier. is it implemented multiple times in the final solution? might be worth finding a way to keep a single copy of the function. one option might be in its own co-located file (e.g. `./nextMeetingSlot.ts`) or with some of its siblings in a local collection of functions (e.g. `./scheduleUtils.ts`).

src/graphql/GetSchedulesForFacility.graphql
```diff
@@ -0,0 +1,11 @@
...
+  $id: ID!
+) {
+  facility(id: $id) {
+    id
+    schedules {
+      id
+      name
+    }
+  }
+}
```
cainlevy, line 11: is this still used? it fetches a surprisingly small amount of data.

src/redux/modules/schedule.ts
```diff
@@ -0,0 +1,47 @@
...
+    const body = await fetchAuthenticated(`schedules`);
+
+    const schedules = (body.data as Record<string, unknown>)
+      .results as Schedule[];
+
+    return schedules;
+  }
+);
+
+export const schedulesSlice = createSlice({
```
cainlevy, line 25: in the end i'm still unsure if we need this redux module. i don't recall seeing any selectors that read from it.

if it's currently essential to the functionality of the pull request, let's keep it and revisit in a follow up. if it's not essential, maybe it's cheap to remove?

cainlevy (approved):
as far as i can tell this is all properly gated by a feature flag which makes reduces the risk of merging it out.

next time, i would encourage taking advantage of the feature flag to merge more smaller branches. large, long-lived branches are harder to maintain (merge) and can sometimes limit access to useful bits of the partial solution.

as it is, though, everything looks good (my comments are mostly administrative) and ready to go from my perspective.

src/redux/modules/schedule.ts
dabbler0: Not essential; I made this to mimic the way we treat kiosks in case it is something we like to use. I'll remove it.

src/graphql/GetSchedulesForFacility.graphql
dabbler0: Seems not, removing.

PR #330, opened 2022-07-14
Sc 4064/meeting requests graphql

crcommons:


src/redux/modules/call.ts
```diff
@@ -1,33 +1,50 @@
...
   "calls/fetchAll",
-  async (filters: MeetingFilters) => {
-    const body = await fetchAuthenticated(
-      `calls?${createCallOptionsParam(filters)}`
-    );
-
-    const calls = (
-      (body.data as Record<string, unknown>).results as CallRO[]
-    ).map(cleanCall) as BaseCall[];
+  async (params: FetchMeetingParams): Promise<BaseCall[]> => {
```
crcommons, line 31: I need to revert these and then create separate graphql-based functions so that I don't mess with other functionality we're not ready to migrate.

crcommons: never mind. looks like it doesn't need to be used at all

src/App.tsx
```diff
@@ -102,20 +105,16 @@ function App({ history }: { history: History }) {
...
           dispatch(
             fetchInPersonVisits({
-              scheduledStart: {
-                rangeStart: startOfMonth(new Date()).getTime(),
-                rangeEnd: endOfMonth(new Date()).getTime(),
-              },
-              limit: 1000,
+              facilityId: selected.id,
+              scheduledStartAfter: startOfMonth(new Date()).getTime(),
+              scheduledStartBefore: endOfMonth(new Date()).getTime(),
```
crcommons, line 110: I'm debating on if it's worth it to move these to use the `useQuery` hook and then add a new redux reducer to hydrate the call/visit data. This works as it is now, but it's following a different pattern. @cainlevy Curious what you think?

cainlevy (approved):

PR #332, opened 2022-07-19
Set Cache-Control header to "no-cache"

FreedomBen:
I don't think this actually does anything since it's the default in most
(all?) browsers, but it's easy to try


deploy/nginx.conf
```diff
@@ -84,6 +84,7 @@ http {
         add_header X-Frame-Options "SAMEORIGIN";
         add_header X-XSS-Protection "1; mode=block";
         add_header X-Content-Type-Options "nosniff";
+        add_header Cache-Control "no-cache";
```
cainlevy, line 87: does this only affect the root path? is there a way to target it?

FreedomBen: No it's going to affect everything.  There is a way to target it to just the root path, but because of the way browsers treat it, it will still affect all the resources linked from that root path (so all the js and css and images loaded by code on the root path)

cainlevy:
I think we should delay this change until we can figure out a more holistic caching strategy that doesn't have the "viral" behavior you've observed before. It seems like a lot of the doubts we had about caching today were better explained by the currently running version of the DOC client.

PR #335, opened 2022-07-19
migrate CallRequestsPage to GraphQL

cainlevy:
This step feels incomplete because accepting or rejecting a meeting does not update the Apollo cache and so the request sticks around. I think it might make sense to import the accept/reject work from https://github.com/Ameelio/connect-doc-client/pull/330 into this branch before releasing it.


src/components/SearchMeetings/SearchMeetingsTable.tsx
```diff
@@ -24,6 +24,16 @@ interface Props {
...
 
+function convertMeeting(m: Meeting): MeetingToReject {
+  return {
+    id: m.id,
+    meetingType:
+      m.type === "call" ? MeetingType.Call : MeetingType.InPersonContact,
+    inmates: m.inmates,
+    visitors: m.contacts,
+  };
+}
```
cainlevy, line 35: The CallRequestsPage is slightly coupled with the SearchMeetingsTable because they both rely on CancelReasonModal. Here, I chose to avoid a SearchMeetingsTable refactor by writing a small conversion function. It's a little bit silly because SearchMeetings queries GraphQL and then converts to Redux types and now also converts _back_ to GraphQL types, but I didn't want to continue refactoring too much at once.

src/components/Requests/CallRequests.tsx
```diff
@@ -120,16 +122,16 @@ export const CallRequests = ({
         title="Kiosk"
         dataIndex="kiosk"
         key="kiosk"
-        render={(kiosk: Kiosk) => (
+        render={(kiosk: MyMeeting["kiosk"]) => (
           <>
-            <span>{kiosk.name}</span>
+            <span>{kiosk?.name}</span>
```
crcommons, line 127: If it's possible for kiosk to be undefined, should we conditionally display this column component?

src/components/SearchMeetings/SearchMeetingsTable.tsx
```diff
@@ -36,7 +46,7 @@ const SearchMeetingsTable: React.FC<Props> = ({
     <>
       {meetingToCancel && (
         <CancelReasonModal
-          entity={meetingToCancel}
+          entity={convertMeeting(meetingToCancel)}
```
crcommons, line 49: I think you could just use the existing `graphQlMeetingToCall` and `graphQlMeetingToVisit`. However, it's all going in the trash eventually, so probably not a big deal

src/pages/CallRequestsPage/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const facility = useSelectedFacility();
+  const { data, loading } = useQuery(GetMeetingRequestsDocument, {
+    onError: errorReporter("GetMeetingRequests"),
+    variables: { facilityId: facility.id },
+  });
+  const meetings = useMemo(() => {
+    if (!data) return [];
+    return data.facility.meetings.edges
+      .map((e) => e.node)
+      .filter((m) => m.status === MeetingStatus.PendingApproval);
```
crcommons, line 28: This filter is unnecessary because we're already filtering for pending meetings on the query itself

src/components/Requests/CallRequests.tsx
cainlevy: It's optional in the type definitions because of voice calls, but should always exist here.

src/components/SearchMeetings/SearchMeetingsTable.tsx
cainlevy: Oh yeah this is converting from a Redux type back to a GraphQL type. Not sure if we already have that conversion somewhere.

crcommons: ah, oops. Getting mixed up with all the gql and redux going on.

cainlevy:
@crcommons i ran through the reject & accept flows and only updated the `fetchPolicy`. i think it's ready to go out tomorrow?

crcommons (approved):

PR #337, opened 2022-07-20
Accept/Reject Meeting Graphql

crcommons:
- Uses graphql to accept and reject meetings
- Updates badge data after status change


src/pages/CallRequestsPage/AcceptMeeting.graphql
```diff
@@ -0,0 +1,31 @@
...
+        }
+      }
+      kiosk {
+        id
+        name
+        description
+        enabled
+        meetingType
+      }
+      call {
```
cainlevy, line 24: Do we still need all of this data back for Redux cache maintenance?

cainlevy (approved):
Does the badge update after accepting/rejecting a request, or is that something we'd want to follow up on?

src/pages/CallRequestsPage/AcceptMeeting.graphql
crcommons: I can probably remove it in my PR where i'm getting rid of useCall. I'll work it out.

PR #338, opened 2022-07-20
Remove useCalls, useIPVisits, and useMeetings

crcommons:
- removes useCalls, useUPVisits, and useMeetings
- creates a liveCalls hook which allows us to remove the other hooks


cainlevy (commented):
it's looking good! i think i'm just uncertain about how to think about cache maintenance. should we refetch periodically, or is there some moment when we merge statuses into the cache and apply a filter, etc.

src/App.tsx
```diff
@@ -44,6 +43,7 @@ function App({ history }: { history: History }) {
   const selected = useAppSelector((state) => state.facilities.selected);
   const pathname = useAppSelector((state) => state.router.location.pathname);
   const dispatch = useAppDispatch();
+  const liveCalls = useLiveCalls(selected, 30000);
```
cainlevy, line 46: @crcommons i moved polling here because App.tsx is responsible for creating a registry of live calls that enables LiveCallPage to work. it's an unfortunate coupling caused by the Iowa DOC request to monitor calls even when not looking at the monitoring page. we essentially maintain the CVH connections at a global level, even if we only consume the streams while looking at the monitoring interface. 😭 

on the plus side, this updates the dashboard and menu badge if calls start after loading the DOC client!

cainlevy (approved):

src/App.tsx
crcommons: Makes sense!

PR #339, opened 2022-07-20
always fetch the latest live calls

cainlevy:
I have no reason to think this has caused problems, but I also don't want it to!


cainlevy:
this backport is no longer needed

PR #340, opened 2022-07-20
clean up meeting related redux modules

cainlevy:
Fixes one remaining `updateCallStatus` thunk and then cleans up the unused meetings / calls / visits Redux logic.


crcommons (approved):
yay!

PR #342, opened 2022-07-20
Fix bug where editing one schedule would carry values over into another

dabbler0:
This was a problem with the old and new pane sharing a a `FormInstance` from `Form.useForm()`, and is fixed by having that `form` instance depend via an effect on the `schedule` parameter.


cainlevy (approved):
what i don't understand is how EditScheduleForm could maintain any state when it's unmounted after `editing` becomes false:

https://github.com/Ameelio/connect-doc-client/blob/9e64c022f17d5721a921220be69662c24e1c19bf/src/components/Settings/EditScheduleDrawer.tsx#L138-L146

dabbler0:
When you close the drawer by clicking outside of it (and therefore don't need to "cancel" or "save"), `editing` was not actually set to false and the drawer wasn't unmounted (just made not `visible`). That's why when you clicked on a new schedule to edit the editor was already up instead of the read-only view. This change also fixes that, so that when the drawer is closed in any way the editor is unmounted.

PR #345, opened 2022-07-22
fix polling-driven reconnect with cached call

cainlevy:
When the DOC polls for the current list of live calls, it gets new authentication tokens. This results in `useConnectCall` deciding to reconnect to CVH, which then causes a half second where the call monitoring is not active and everyone's video freezes.

My fix is to cache the call object in the `<RegisterCall />` component so that changes to the `call` prop will not cascade down to `useConnectCall`.


jackofsometrade (approved):
:+1:

crcommons (approved):
This makes sense to me. I can't see how this would have negative effects downstream.

PR #350, opened 2022-07-25
Reject Modal Styling Updates

crcommons:
https://user-images.githubusercontent.com/8941223/180840585-47e45c02-38e2-4079-87e8-590605fcfa56.mov


cainlevy (approved):
wow that looks so much better

PR #353, opened 2022-07-25
remove inmate quota editing

cainlevy:
We've settled on managing this information via roster updates. Any edits made here would be lost in the next roster.


src/App.tsx
```diff
@@ -85,7 +84,6 @@ function App({ history }: { history: History }) {
         await Promise.allSettled([
           dispatch(fetchContacts()),
           dispatch(fetchStaff()),
-          dispatch(fetchInmates()),
```
cainlevy, line 88: nothing was reading from the redux slice that this populated. the last selector was unused.

crcommons: ✂️

crcommons (approved):
So happy to see more stuff getting cut

PR #360, opened 2022-07-28
Column Width Fix

crcommons:


src/components/Requests/CallRequests.tsx
```diff
@@ -38,6 +39,7 @@ export const CallRequests = ({
   return (
     <Table dataSource={meetings} loading={loading}>
       <Table.Column
+        className="date-time-column"
```
crcommons, line 42: You can't do inline styling with this ANT Design component, so classes seem to be the best option

PR #364, opened 2022-08-05
Sc 4021/directory of partnered organizations

crcommons:
https://user-images.githubusercontent.com/8941223/183047129-96876b3f-4cdd-46ae-a0e1-0e5153525136.mov


src/pages/OrganizationsPage.tsx
```diff
@@ -0,0 +1,188 @@
...
+const { Option } = Select;
+
+type FacilityOrganization =
+  GetPartnerOrganizationsQuery["facility"]["facilityOrganizations"][0];
+
+type Organization =
+  GetPartnerOrganizationsQuery["facility"]["facilityOrganizations"][0]["organization"];
+export default function OrganizationsPage(): ReactElement {
+  const facility = useSelectedFacility();
+  // const navigate = useAppNavigate();
```
crcommons, line 37: will re-enable on next ticket

src/pages/OrganizationsPage.tsx
```diff
@@ -0,0 +1,188 @@
...
+              <Typography.Text>{formatDate(date, "dateyear")}</Typography.Text>
+            )}
+          />
+
+          <Column
+            title=""
+            key="view"
+            render={(_text, facilityOrg: FacilityOrganization) => (
+              <Button
+                // onClick={() => navigate(`/organizations/${facilityOrg.id}`)}
```
crcommons, line 177: will re-enable on next ticket

src/constants/Routes.ts
```diff
@@ -108,6 +109,11 @@ export const ROUTES: Route[] = [
   },
   // index should be last entry
   { path: "/", component: DashboardPage, label: "Dashboard" },
+  {
+    path: "/organizations",
+    component: OrganizationsPage,
+    label: "Partner Organizations",
+  },
```
cainlevy, line 116: does this prove the "index should be last entry" comment wrong?

src/pages/OrganizationsPage.tsx
```diff
@@ -0,0 +1,188 @@
...
+  });
+
+  useEffect(
+    () =>
+      data &&
+      setFacilityOrgs(
+        data.facility.facilityOrganizations.filter(
+          (fo) =>
+            fo.organization.name.toLowerCase().includes(query.toLowerCase()) ||
+            fo.organization.type.toLowerCase() === query.toLowerCase()
```
cainlevy, line 77: makes me wonder if we should have an omnisearch and forego making the user declare "name" or "type". it's probably a discussion for design on another day.

src/pages/OrganizationsPage.tsx
```diff
@@ -0,0 +1,188 @@
...
+          <Column
+            title="Name"
+            dataIndex={["organization", "name"]}
+            key="organization.name"
+          />
+          <Column
+            title="Type"
+            dataIndex={["organization"]}
+            key="organization"
+            render={(org: Organization) => (
```
cainlevy, line 149: i've started referring back to the dataSource's type:

```suggestion
            render={(org: FacilityOrganization['organization']) => (
```

or 

```suggestion
            render={(org: typeof facilityOrgs['organization']) => (
```

mostly because i like how it matches up with the `dataIndex` prop.

just an idea.

cainlevy (approved):

src/pages/OrganizationsPage.tsx
crcommons: I considered posing this question to design so we could just make the interface a little more generic. But probably worth a discussion in the next providers meeting. Will add to the backlog

src/constants/Routes.ts
crcommons: Somehow I glossed over that comment. Doesn't seem like it mattered, but I changed it just in case.

PR #366, opened 2022-08-09
Organization details page

crcommons:
https://user-images.githubusercontent.com/8941223/183781547-e5a11f3d-ecf0-4ddc-b38a-a8e889d514e3.mov


src/Organizations/OrganizationDetails.tsx
```diff
@@ -0,0 +1,138 @@
...
+  | "email"
+  | "phone"
+> & {
+  admin: Pick<Visitor, "firstName" | "lastName" | "email">;
+};
+
+export default function OrganizationDetails({ admin, ...organization }: Props) {
+  return (
+    <>
+      <Space direction="vertical" size="middle" style={{ width: "100%" }}>
```
crcommons, line 30: Decided that I would swap out the previous implementation of the organization details for one that matches the column/row pattern I've seen throughout that app. So 99% of this component is the same as `OrganizationRequestDetails` but just uses rows and columns instead of items

src/graphql/GetFacilityOrganization.graphql
```diff
@@ -0,0 +1,26 @@
...
+        }
+      }
+      memberships {
+        id
+        visitor {
+          id
+          firstName
+          lastName
+          email
+          profileImageUrl
```
cainlevy, line 19: VisitorFragment?

cainlevy (approved):

PR #367, opened 2022-08-10
Connection Request Await Review and Cache policy 

crcommons:
2 fixes here:

- One issue was that we need to await the accept/reject mutation and disable the buttons while waiting so that there isn't confusion for the user while the request completes
- While doing that, I found an issue with the cache policy in that component which was causing an INFINITE RERENDER!!!! 💀  Implemented a fix - More info can be found here https://github.com/apollographql/apollo-client/issues/6819


cainlevy (approved):
oh i'm having some flashbacks on this bug. let's see how it develops.

PR #370, opened 2022-08-12
Enable providers features for superusers

dabbler0:
(If this is in fact the way we want to go for this).


src/components/Settings/ScheduleSettings.tsx
```diff
@@ -29,6 +30,9 @@ const ScheduleSettings: React.FC<Props> = () => {
   const facility = useSelectedFacility();
   const role = useSelectedFacilityRole();
 
+  // Can remove the following once providers is enabled
+  const session = useAppSelector((state) => state.session);
```
cainlevy, line 34: looks unused

cainlevy (approved):
I think this unblocks us now and does not interfere with a facility feature later

PR #371, opened 2022-08-16
shorten menu text to not overflow-hide badge

wittejm:
This is my poor-person's solution to overflow, and I'm stopping after _just 30 minutes_ of fiddling with CSS. I _could_ spend the rest of the day fiddling to get a flex-based solution to work or something. Or just stick with this. Or if you know a "principled" approach to suggest here, I could put that in. 

Screen Shot 2022-08-16 at 11 39 36 AM


src/components/Menu/Menu.tsx
```diff
@@ -275,7 +275,8 @@ export default function Menu({
             key="professional-requests"
             title={
               <span>
-                Professional requests{" "}
+                Professional reque
+                {allProfessionalPendingRequestsCount > 0 ? "" : "sts"}
```
cainlevy, line 279: 😂

this is kind of amazing. we don't have to support translation or variable menu sizes so sure, why not!

```suggestion
                {allProfessionalPendingRequestsCount > 0 ? "&nbsp;…" : "sts"}
```

cainlevy (approved):

PR #373, opened 2022-08-16
remove unused kiosk.groups references

cainlevy:
The kiosk.groups relationship is no longer maintained and needs to be cleaned up.


src/api/graphql.ts
```diff
@@ -741,7 +741,6 @@ export type Kiosk = {
   description: Scalars["String"];
   enabled: Scalars["Boolean"];
   facility: Facility;
-  groups: Array<Group>;
```
cainlevy, line 744: i tested removing this from the backend to see what would complain in the frontend. that's how i found the unused code.

crcommons (approved):
Yay! love cleanup!

PR #375, opened 2022-08-18
Feature flag for system management

wittejm:
First bit of design indecision: do we navigate to the two views via a single button in the menu? or two? If it's one button, what will navigation between the two views look like? Favorite solution: a submenu!
Screen Shot 2022-08-18 at 3 48 47 PM


.github/workflows/build-test-deploy.yml
```diff
@@ -52,6 +53,7 @@ jobs:
           REACT_APP_API_URL: "https://connect-api.ameelio.org/api/v1/"
           REACT_APP_SENTRY_DSN: "https://af36ac2171b4413da95f553d5bdfc5df@o434922.ingest.sentry.io/6021906"
           REACT_APP_ENABLE_PROVIDERS: ""
+          REACT_APP_ENABLE_SYSTEM_MANAGEMENT: ""
```
cainlevy, line 56: this has a few more steps than you'd guess: https://github.com/Ameelio/connect-doc-client#adding-an-environment-variable

src/components/Menu/Menu.tsx
```diff
@@ -172,6 +172,11 @@ export default function Menu({
           </AntdMenu.Item>
         ))}
       </SubMenu>
+      {featureEnabled("SYSTEM_MANAGEMENT") && (
+        <AntdMenu.Item onClick={() => window.alert("system management!")}>
+          <span>System Management</span>
+        </AntdMenu.Item>
```
cainlevy, line 178: should this include the submenu?

i do think submenus are most consistent with the doc client's navigation setup.

cainlevy (changes requested):

src/components/Menu/Menu.tsx
wittejm: yep I forgot to push my last commit :/

jackofsometrade:
Sub menu looks good to me!  👍

cainlevy (approved):

jackofsometrade (approved):

PR #377, opened 2022-08-18
headers design updates

cainlevy:
One of the last bits of the Providers epic is a redesign and touchup to related pages in the DOC client. The design called for updating the header on the Resident page, but I went ahead and applied the basic style change to the rest:

image
image
image
image
image


src/components/Profile/ContactIDButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+import React, { useState } from "react";
+import { Visitor } from "src/api/graphql";
+import { isMinor } from "src/utils";
+import VisitorPhotosModal from "./VisitorPhotosModal";
+
+type Props = {
+  visitor: Pick<Visitor, "id" | "firstName" | "lastName" | "dateOfBirth">;
+};
+
+export default function ContactIDButton({ visitor }: Props) {
```
cainlevy, line 12: extracted to reduce the complexity of page header implementation on the contact page

src/pages/InmatesPage.tsx
```diff
@@ -82,8 +82,8 @@ const InmatesPage: React.FC = () => {
         <ImportModal onClose={() => setShowImportModal(false)} />
       )}
       <Header
-        title="Incarcerated People"
```
wittejm, line 85: I'm generally giving way on my anti-"resident" language stance but in this case can we stick with Incarcerated People? To remind our users that they're providing prison, not housing?

src/pages/InmatesPage.tsx
```diff
@@ -82,8 +82,8 @@ const InmatesPage: React.FC = () => {
...
-        subtitle="Manage the members of your facility, access detailed information, and edit their info as needed."
+        title="Residents"
+        subtitle="Manage the residents of your facility, access detailed information, and edit their info as needed."
```
wittejm, line 86: "members" is ridiculous though. This is fine.

wittejm (approved):

src/pages/InmatesPage.tsx
cainlevy: I hear your reasoning but I'm not sure who can settle the debate once and for all.

At present, I'm aiming for consistency with the navigation menu which says "Directory > Residents".

src/pages/InmatesPage.tsx
cainlevy: The DOC client probably still contains every single word choice we've tried, somewhere.

src/pages/InmatesPage.tsx
wittejm: Yeah, I'm not sure it can be settled and that's fine. In this case I push for a bit more clarity of meaning over consistency of UX.

wittejm: We can have another whole team debate about it! Wanna?

jackofsometrade: I would vote for more precise language in this case.  I don't see a need for delicate language considering the users are prison staff.  Are there any possible futures in which the use of the word "incarcerated" would be incorrect for the facility type?

cainlevy: I don't see any such futures, but I also don't think this is the appropriate place to make a decision. More specifically, I don't think I can take ownership of the decision.

cainlevy: I don't really have a horse in this race, except that I think it's important for navigation labels and page titles to match.

src/pages/ContactPage.tsx
```diff
@@ -29,16 +30,57 @@ function ContactPage({ match }: RouteComponentProps<TParams>): ReactElement {
...
+            <ContactIDButton visitor={visitor} />
+          </div>,
+        ]}
+        breadcrumb={{
+          routes: [
+            { path: "/contacts", breadcrumbName: "Contacts" },
+            { path: `/${visitor.id}`, breadcrumbName: genFullName(visitor) },
+          ],
+        }}
+        onBack={() => window.history.back()}
```
jackofsometrade, line 53: I was looking for a way to `.back()` or `(-1)` on `useAppNavigate()` but that doesn't seem to exist 😢

jackofsometrade (approved):

src/pages/InmatesPage.tsx
eliz214: agree with what's been said here in that clarity is what we should shoot for + consistent language across nav and tabs, what i think could be most useful (I'm not sure if this is possible) is if we could make that language a variable that changes based on the language of the facility, as this can really vary. some places say AICs (adults in custody), residents, IIs (incarcerated individuals), offenders, inmates.... that's just what i remember from the top of my head right now.

eliz214:
I'm so excited about these changes! The header's been needing a facelift for a while. Three small requests now that I'm seeing all the variations:

- Can we align the icon + text combos for buttons (i.e. See Photo ID) and nav? 
- I mentioned this in the last thread, but let's standardize the spacing of the content below the header. Right now, the horizontal spacing looks thinner than the vertical spacing from the header. We should reduce the vertical to match the horizontal.
- The alignment of some of the search components is a bit wonky. Some lines are next to each other, some look on top of each other, and the icon is a bit askew. Can we update this? Lmk if you need a design for clarity
image

cainlevy:
1. ✅ (finally!)
2. ✅ 
3. sorry 😞 i think this is another undesired interaction of our hybrid design system (antd + tailwind) and i couldn't spot an easy fix.

PR #378, opened 2022-08-18
redesign activity and connection cards for visitors and residents

cainlevy:
The redesign switches tabs for dropdown filters and streamlines the activity table for the available space.

image

image


wittejm (approved):

eliz214:
looking good! a few nit requests:

- can we standardize the margins around the activity and connection cards? right now, the horizontal margin is smaller than the vertical. let's update the vertical to equal the horizontal
- can we update the dropdown style to the text variant?
- 'See details' button should use the small variant. Can we update the status components to be in the style shown? Right now the styles of the status tag and the button are competing visually for me
image
- here's the [figma frame](https://www.figma.com/file/Hyz93bxJuWSrIuSkHA2mBf/DOC%E2%80%94Provider-updates?node-id=2%3A15373) again for specific reference

cainlevy:
@eliz214 should the status component use that dot style everywhere? i'm currently using a colored tag component that also shows up in other places.

cainlevy:
@eliz214 it feels like the colored tag pattern has become a core design element in the doc client. i wonder if this particular one is just missing the color or if there's some other choice we can make to more clearly differentiate between tags and buttons.

cainlevy:
@eliz214 Jason suggested trying a pill shape for DOC client tags. I'm happy to revisit when you return. Here's an updated screenshot:

image

PR #379, opened 2022-08-19
move components around a bit

cainlevy:
I keep wanting to move components around while I'm working on other branches but then I don't because it would add confusion to the pull request.

So here's the shuffle. I tried to:

- keep page specific concerns with the page
- avoid organizing by type
- flatten some unnecessary subfolders
- remove some dead code


jackofsometrade (approved):
I considered questioning all the moving arounds, but I trust you have good reasons for doing so, and everything looks fine to me at a glance.

Edit: and now I see your list 👍 😄

PR #383, opened 2022-08-22
sc-4443 system staff view skeleton (+ updated system facilities view)

jackofsometrade:
https://user-images.githubusercontent.com/97806603/186031847-d3f3a94c-f1bd-4bfe-b1c0-d6e1e0fe31d1.mp4


jackofsometrade:
Added SystemPage as a skeleton for root level things if needed, and to provide sense to the breadcrumbs in facilities and staff pages.

src/pages/SystemPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+            {
+              path: "/system",
+              breadcrumbName: `${data.facility.system.name} DOC`,
+            },
+          ],
+        }}
+        onBack={() => window.history.back()}
+      ></PageHeader>
+      <div style={WRAPPER_STYLE}>
+        <Link to={`/system/facilities`}>Facilities</Link>
```
wittejm, line 42: I notice we don't use this `Link` anywhere in doc client but instead use antd's `Typography.Link` for in-site navigation. Let's stick with that?

src/graphql/GetSystemStaff.graphql
```diff
@@ -0,0 +1,22 @@
+query GetSystemStaff($facilityId: ID!) {
+  facility(id: $facilityId) {
+    system {
```
wittejm, line 3: You had mentioned the idea of adding `staff` directly to `System` in the schema/resolver. Do you still think we should do that at some point?

wittejm (approved):

src/pages/SystemPage.tsx
jackofsometrade: Oh ok, sure thing

src/graphql/GetSystemStaff.graphql
jackofsometrade: I do.  I was thinking we might do that if and when we need pagination?

cainlevy, line 3: might want to ensure the query has `id` fields everywhere for proper caching

PR #384, opened 2022-08-23
sc-4440 facility features column

jackofsometrade:
Going off a fairly brief ticket with no visual design, please don't be shy with your feedback 🙏 

https://user-images.githubusercontent.com/97806603/186240735-45eb4b72-6055-4049-a027-1e0adb8fa047.mp4


jackofsometrade:
@wittejm I hope I interpreted the requirements properly, LMK if anything is off.

cainlevy (commented):
@cshdllr we have an assortment of other possibly interesting configuration details that may be interesting at a summary level. as you will see, there is little symmetry to them:

## messaging

- enabled/disabled

## voice calls

- enabled/disabled
- maximum duration of a single voice call
- maximum number of concurrent voice calls
- voice call time quota per resident

## video calls

- enabled by facility schedule configuration
- video calls per week per resident (unless changed with roster upload)
- whether video calls require monitoring (unless changed with roster upload)
- whether meetings (including video calls) require approval (unless changed with roster upload)
- scheduling window for meetings (including video calls)

## visits

- enabled by facility schedule configuration
- in person visits per week per resident (unless changed with roster upload)
- whether meetings (including visits) require approval (unless changed with roster upload)

cshdllr:
@cainlevy do we want to include that information here? Or are you just mentioning that in case we do want to include it here? 

Will there be a page with more detail for a specific facility (ex. Test Facility)? That could be a nice place to add that information.

cainlevy:
I was cataloguing it just in case.

My intuition is that we'd get more value from a form than a details page, but I don't know if that should be part of a bigger facility settings revamp or where it would really end up.

wittejm (approved):

cshdllr (approved):

PR #385, opened 2022-08-23
Add Facility modal

wittejm:
This copies much of the code from the existing Add Facility form on the [Ameelio panel ](https://github.com/Ameelio/connect-doc-client/blob/main/src/components/Admin/EditFacility.tsx) But it:

- Doesn't clear the form if the request fails
- Shows validation error messages returned by graphql. There's only one field with a unique constraint, so we rest on that assumption when toasting the "unique constraint violation" error./
- is on the new System Facilities page and updates the contents of that page. 

https://user-images.githubusercontent.com/2104990/186274078-6ea711b7-7af1-472f-bc98-5d1f6d867a61.mov


src/components/System/AddFacilityModal.tsx
```diff
@@ -0,0 +1,293 @@
...
+            rules={[
+              {
+                required: formItem.required,
+                message: "",
+              },
+            ]}
+          >
+            {getInputBox(formItem)}
+          </Form.Item>
+        ))}
```
cainlevy, line 281: i was curious what the payoff for configuring a `formItems` array would look like. seems like it's not much. apart from the `rules` construction, i'd guess we'd end up with net neutral (or even net negative) LOC by inlining all of it.

src/pages/SystemFacilitiesPage.tsx
```diff
@@ -29,12 +35,14 @@ const SystemFacilitiesPage = () => {
...
-              Add Facility
-            </Button>
+            {isSuperUser(session.user) && (
+              <Button
+                type="primary"
+                onClick={() => setIsAddFacilityModalOpen(true)}
+              >
+                Add Facility
+              </Button>
+            )}
```
cainlevy, line 45: yeah this is a button that i think is better off hidden from normal users

cainlevy (approved):

src/pages/SystemFacilitiesPage.tsx
wittejm: Yep. I think the general reason to not hide buttons is so users don't get confused as to why it's missing (I've had that trouble before with connection restriction features being hidden to non-admins). But Ameelio superusers will hopefully not be facing this confusion.

PR #387, opened 2022-08-24
Use simpler system.staff query on system staff page

wittejm:
This also deduplicates entries for staff with multiple staff positions.


cainlevy (approved):
🎉

PR #392, opened 2022-08-25
clear cache on logout

cainlevy:
Ty noticed that DOC clients did not disconnect from calls on logout. The reason is because `App.tsx` manages both authentication concerns and the CallRegistry, and we were holding onto the entire Apollo cache (including the list of live calls). The result is that CallRegistry didn't know anything had changed and continued to hold onto the call connections.

The simple solution is to clear the Apollo cache. This should happen regardless, and it also fixes this case by removing the data that CallRegistry observes, triggering disconnections.


src/App.tsx
```diff
@@ -72,6 +72,12 @@ function App({ history }: { history: History }) {
       dispatch(setRedirectUrl(pathname));
   }, [dispatch, isAuthenticated, pathname]);
 
+  // on logout, clear apollo cache
+  const client = useApolloClient();
+  useEffect(() => {
+    if (!isAuthenticated) client.clearStore();
+  }, [isAuthenticated, client]);
```
cainlevy, line 79: @jackofsometrade fixed one of the monitoring bugs!

wittejm (approved):

PR #393, opened 2022-08-25
Migrate login flow to graphql

dabbler0:
This is in preparation for including entitlements information in the session in redux.


src/redux/modules/session.ts
```diff
@@ -118,9 +121,16 @@ export function sessionReducer(
...
       return {
         ...state,
-        user,
-        authInfo: { token, type: "doc", id: user.id },
-        status: "active",
+        user: user || state.user,
+        authInfo:
+          token && user
+            ? { token: token, type: "doc", id: user.id }
+            : state.authInfo,
```
cainlevy, line 128: is this still useful?

src/pages/LoginPage/index.tsx
```diff
@@ -9,38 +10,111 @@ import {
...
+      input: {
+        username: email,
+        password,
+        userType: UserType.Staff,
+      },
+    },
+    onError: () => {
+      showToast(TOAST_KEY, "Invalid email or password", "error");
+    },
+    skip: !email || !password,
```
cainlevy, line 54: ahh nice!

another option is `useLazyQuery` which gives you an `await`-able function, much like `useMutation`. then you could trigger it directly from `onFinish`.

src/pages/LoginPage/index.tsx
```diff
@@ -9,38 +10,111 @@ import {
...
+  });
+
+  // Set auth token once login query finishes
+  useEffect(() => {
```
cainlevy, line 58: is there an `onCompleted` hook available? which do you prefer?

src/pages/LoginPage/index.tsx
```diff
@@ -9,38 +10,111 @@ import {
...
+          token: loginData.loginUser.authToken,
+        })
+      );
+    }
+  }, [loginData]);
+
+  // Current user fetch query which will run once login query is done
+  const { data: currentStaffData } = useQuery(CurrentStaffDocument, {
+    onError: errorReporter("currentStaff"),
+    skip: session.token === "",
```
cainlevy, line 71: could also try `useLazyQuery` for normal await-based sequencing

src/pages/LoginPage/index.tsx
```diff
@@ -9,38 +10,111 @@ import {
...
-      });
-    } catch (err) {
-      showToast(TOAST_KEY, "Invalid email or password", "error");
-    }
+
+    setEmail(email);
+    setPassword(password);
   };
 
   const onFinishFailed = (_errorInfo: any) => {
```
cainlevy, line 120: might not be useful if `onFinish` reports its own errors?

cainlevy (approved):
looks good, i think all of my comments are just additional options

src/redux/modules/session.ts
dabbler0: It's still used in `ConnectCall` and possibly some other places; it could of course be reassembled there from `user` and `token`.

src/pages/LoginPage/index.tsx
dabbler0: I think I'll try this to see if it gets rid of redux's warning about changing state during render.

src/redux/modules/session.ts
cainlevy: got it, sounds like something that can happen later

src/pages/LoginPage/index.tsx
dabbler0: Hmm, it seems that in the version of apollo client we're using here, the `await`-able function from `useLazyQuery` returns `void` as opposed to the actual results (it just triggers the query and then you need to get the results from the hook itself). So for the sequencing we want we'd need rerenders anyway. I'll keep things as they are for now.

PR #394, opened 2022-08-25
Disable buttons if you don't have the right entitlement

dabbler0:
Finally finished after a bunch of design iterations. Disables navbar items that you're not allowed to access and puts up a "locked" splash screen if you manually enter those urls anyway. For pages that are still readable become write-only, the write buttons are made disabled and given a tooltip, with few exceptions. Those exceptions are in modals (where disabling "Cancel/Ok" seems confusing; they are instead removed so that the "x" button is salient) and in the Schedule Settings page where "View/Edit Details" is changed to "View Details" when read-only.


wittejm:
would a video screencap be useful for showing any major design decisions to get feedback from Liz or Jason?

src/App.tsx
```diff
@@ -39,6 +40,9 @@ function App({ history }: { history: History }) {
   const dispatch = useAppDispatch();
   const liveCalls = useLiveCalls(selected, 30000);
   const isAuthenticated = session.status === "active";
+  const canMonitorLiveCalls =
```
wittejm, line 43: looks like this is no longer used in this file?

src/pages/MessageReviewPage.tsx
```diff
@@ -194,44 +202,48 @@ const MessageReviewPage = (props: Props) => {
...
-        }
-        onReject={(reason, internalComments) =>
-          flaggedMessage &&
-          rejectMessage({
-            variables: {
-              input: {
-                messageId: flaggedMessage.id,
-                statusDetails: reason,
-                internalComments,
+      {canMonitorLiveCalls ? (
```
wittejm, line 205: wasn't there the idea of splitting out message review to its own entitlement?

src/pages/MessageReviewPage.tsx
```diff
@@ -194,44 +202,48 @@ const MessageReviewPage = (props: Props) => {
...
-        }
-        loading={rejectLoading || approveLoading}
-      />
+              update: (cache) => decrementPendingMessagesCount(cache),
+            })
+          }
+          loading={rejectLoading || approveLoading}
+        />
+      ) : (
+        <NotAllowed />
```
wittejm, line 245: same comment as above about the long ternary. I won't tag any more examples if I see them :)

src/pages/StaffRosterPage.tsx
```diff
@@ -114,7 +114,7 @@ export default function StaffPage(): ReactElement {
               </>
             )}
           />
-          {isUserAdmin && (
+          {canManageFacility && (
```
wittejm, line 117: shouldn't this be `canManageStaff`?

wittejm (approved):

src/pages/StaffRosterPage.tsx
dabbler0: Good catch; this is also a place where I forgot to change the old "omitting" behavior to "disabled" behavior. Fixing both atonce.

src/pages/MessageReviewPage.tsx
dabbler0: Yep, forgot about that. Changed now, although this means we can't merge this doc client until after the relevant db migration is done in the backend.

src/components/NotAllowed.tsx
```diff
@@ -0,0 +1,15 @@
+import { LockOutlined } from "@ant-design/icons";
+import { Result } from "antd";
+import React from "react";
+
+const NotAllowed: React.FC = () => {
```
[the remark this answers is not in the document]
dabbler0: Yep, these should only be reachable by manually typing the url.

src/pages/LiveCallPage/index.tsx
```diff
@@ -113,181 +121,192 @@ const LiveVisitationContainer: React.FC = () => {
...
+                          className="mt-4"
+                        />
+                      ))}
+                  </div>
+                </div>
+              )}
+            </Sider>
+          </>
+        ) : (
+          <NotAllowed />
```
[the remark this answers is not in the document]
dabbler0: Rearranged all the ternaries that involve `<NotAllowed />` to put `<NotAllowed />` first.

wittejm:
🍾

PR #397, opened 2022-08-26
prefer session.token in fetchAuthenticated

cainlevy:
`fetchAuthenticated` is only used for staff management and roster uploads. Since staff management is so rare, it only showed up for people uploading rosters.


src/api/Common.ts
```diff
@@ -49,7 +49,7 @@ export async function fetchAuthenticated(
       headers: {
         Accept: "application/json",
         "Content-Type": "application/json",
-        Authorization: state.session.authInfo.token,
+        Authorization: `Bearer ${state.session.token}`,
```
cainlevy, line 52: `authInfo` wasn't being set because the reducer action was being called with the token separately from the user. it seems this is only impactful here.

dabbler0 (approved):

PR #398, opened 2022-08-30
detailed errors & lint

cainlevy:
Show the actual error message unless it's "Internal Server Error". This is useful when people fail validations and need to understand why.

Also cleaned up some other lint:

- the warning about updating state during a render on login
- the warning about merging facility data on dashboard
- some missing react keys (antd patterns seem really prone to this)


cainlevy:
[sc-4567]

jackofsometrade (approved):
L👀ks good to me

PR #406, opened 2022-09-01
various system page polish issues

wittejm:
- breadcrumbs should say "Dashboard" and "System Management; and also "Settings"
- In nav menu rename both links “Switch Facility” and “Manage System”
- On the facilities page, The “import roster” button should be primary, not the Add Facility button
- Change Tab names & table header names to “All Facilities” and “All Staff”
- Alignment of things to the left: page header, tab text, and table edge
- Make facility feature tags the default gray color
- “Open Settings” is a link instead of the Settings button
- For staff: “View/Edit Details” instead of Manage button
- When we go back to the dashboard, toast: “now viewing [facility x] dashboard”.


dabbler0 (approved):
Lgtm, other than having the breadcrumbs say "Dashboard" instead of "Exit..." (I remember discussing these options with Liz at once point and feeling like having "Dashboard" be the root feels weird because it implies that the facility-specific dashboard is somehow the parent of the system-wide view, when logically it's the other way around.) But this seems like a minor design decision that we can revisit later, so approving for now and @eliz214 can weigh in on how the result feels.

eliz214:
@dabbler0 — @wittejm and i discussed this very briefly when we were talking about this list of changes. after seeing everything in hifi, i think 'Dashboard' is the best option to summarize going back to the original flow, especially given the toast we're adding that indicates what the new view is. the only other option i could reasonably think of is 'Home' but i feel meh about that

PR #407, opened 2022-09-06
Delete staff button

dabbler0:
Depends on https://github.com/Ameelio/connect-api-nest/pull/1158. Adds a "delete" button the edit staff modal which deactivates staff. Also filters out deactivated staff (i.e. staff with no permissions) so that they don't appear in the system staff list.


src/components/System/AddEditStaffModal.tsx
```diff
@@ -60,9 +88,28 @@ const AddEditStaffModal = ({ onClose, staff, systemId }: Props) => {
...
       }
-      onOk={form.submit}
-      okText="Save"
+      footer={[
+        <Button key="back" onClick={onClose}>
+          Cancel
+        </Button>,
+        <Button key="delete" onClick={() => setDeleting(true)} danger>
+          Delete
+        </Button>,
```
cainlevy, line 97: i think the modal can also be used for adding staff. if so, we should hide this button until `staff` exists.

cainlevy (approved):

PR #411, opened 2022-09-08
use pendingMessageCount field instead of shielded pendingMessages.count

wittejm:
requires https://github.com/Ameelio/connect-api-nest/pull/1165


wittejm:
https://github.com/Ameelio/connect-api-nest/pull/1166

PR #412, opened 2022-09-08
clean up old facility and staff admin tools

cainlevy:
1. Remove the old forms to add facilities and staff
2. Remove the ability to delete staff from the facility-level Directory > Staff page
3. Clean up unused typings


wittejm (approved):
🎉 🎉 🎉

PR #413, opened 2022-09-08
disconnect react router from redux

cainlevy:
We were using connected-react-router, which integrates react-router with redux. This changeset disconnects from redux in favor of the underlying react-router@5.2.0.

Not included:

- switching from HashRouter to BrowserRouter
- updating to react-router@6


src/hooks/useAppNavigate.ts
```diff
@@ -1,16 +1,15 @@
-import { push } from "connected-react-router";
 import { useCallback } from "react";
+import { useHistory } from "react-router";
 import { Path } from "src/constants";
-import { useAppDispatch } from "src/redux";
 
 export function useAppNavigate(): (path: Path) => void {
```
cainlevy, line 5: at this point `useAppNavigate` appears to be mostly useful as a way to add type safety. that's still kind of interesting to me!

crcommons (approved):
woo! \o/

PR #414, opened 2022-09-08
show deactivated staff at bottom of page, and allow re-activation

wittejm:
https://user-images.githubusercontent.com/2104990/189239156-bcfa838d-61d8-491d-99cc-a014eb773206.mov


wittejm:
@cainlevy bump

src/components/System/SystemStaff.tsx
```diff
@@ -101,6 +106,15 @@ const SystemStaff = () => {
...
         </Card>
+        <Collapse>
+          <Panel header="Deactivated Staff" key="deactivate-panel">
+            <Table
+              dataSource={deactivatedStaff || []}
+              key="id"
+              columns={columns}
+            ></Table>
+          </Panel>
+        </Collapse>
```
cainlevy, line 117: maybe hide this when there are no deactivated staff?

src/components/System/SystemStaff.tsx
```diff
@@ -29,6 +31,9 @@ const SystemStaff = () => {
   const staff = data
     ? data.facility.system.staff.filter((s) => s.staffPositions.length > 0)
     : [];
+  const deactivatedStaff = data
+    ? data.facility.system.staff.filter((s) => s.staffPositions.length === 0)
+    : [];
```
cainlevy, line 36: does this still work on a page refresh after all of the positions are inactive?

wittejm: ....hard to say? If I deactivate all staff positions, I no longer have any ability to view the page! I also can't deactivate my own staff account. 

What use case are you imagining here?

cainlevy: We're aiming to support a use case where someone rejoins the DOC after a few months and needs access again. Currently this special case has to go through support.

wittejm: yes, this behavior supports that use case. I'm still confused about what you're asking here: 

> does this still work on a page refresh after all of the positions are inactive?

But: I can deactivate an account, which sets all the positions to inactive, and then I can re-activate it, and that staff account appears in each of the two lists according to that inactive/re-activated state.

cainlevy (approved):

PR #417, opened 2022-09-09
logout upon login if no active staff positions

wittejm:
https://user-images.githubusercontent.com/2104990/189442895-2d27ac10-89a0-45ce-bc5a-a9e6c99c1bee.mov


cainlevy (approved):
yeah that looks good!

PR #419, opened 2022-09-14
Jw/new object types in staff timeline

wittejm:
this is to fix the timeline view so I can proceed with supporting `system` on timeline event.  requires https://github.com/Ameelio/connect-api-nest/pull/1176


cainlevy (approved):
was the timeline view erroring without this new data? or was the server erroring trying to return the new data?

wittejm:
The resolver was returning `null` for these object types and the client got mad at seeing `null` in a non-nullable field.

PR #420, opened 2022-09-15
use systemId to query system data

wittejm:
I needed to extend the redux type to include `systemId` to support this.


src/App.tsx
```diff
@@ -52,8 +53,12 @@ function App() {
     skip: !isAuthenticated,
     onError: errorReporter("App/facilities"),
   });
-  const facilities = useMemo(
-    () => data?.currentStaff?.staffPositions.map((sp) => sp.facility),
+  const facilities: Facility[] | undefined = useMemo(
```
wittejm, line 56: building an ad hod object so this yelled at me without a type signature.

cainlevy (approved):

PR #423, opened 2022-09-16
fix peer dependencies for npm v8

cainlevy:
NPM v8 will not resolve conflicting peer dependencies for production dependencies. This started blocking deploys yesterday. This branch reproduces the problem by updating the local .node-version to 16.16 and then upgrades:

- `@apollo/client` for compatibility with newer `graphql` libraries
- `react-chartjs-2` for compatibility with react@17
- `html-react-parser` for compatibility with react@17
- `react-redux` for compatibility with react@17

There are still peer dependency conflicts in dev dependencies. I plan to finish updating dependencies today.


src/hooks/useLiveCalls.ts
```diff
@@ -23,7 +23,7 @@ export default function useLiveCalls(
   const { data } = useQuery(GetLiveCallsDocument, {
     onError: errorReporter("GetLiveCalls"),
     variables: {
-      facilityId: facility?.id,
+      facilityId: facility?.id || "",
```
cainlevy, line 26: apollo doesn't allow undefined anymore, so we have to provide placeholder values when variables are used in the `skip`

wittejm (approved):

PR #424, opened 2022-09-16
just npm run graphql

wittejm:
no other changes.


wittejm:
Nope nevermind I was running this against an outdate api

PR #425, opened 2022-09-16
facility and entitlement columns and filters on staff page

wittejm:
also extract new `MultiSelect` component and `prettyLabel` util

https://user-images.githubusercontent.com/2104990/190725188-caa87ac8-a6b0-4789-b4fe-f4384b6ef26d.mov


wittejm:
oh darn editing staff facilities/entitlements no longer is immediately reflected in the table. fixing....

wittejm:
Ok fixed. In this case I'm confused why triggering the useEffect on data works while triggering useMemo on data which updates staff which should trigger the useEffect.... doesn't work.  🤷

src/components/System/SystemStaff.tsx
```diff
@@ -28,15 +52,37 @@ const SystemStaff = () => {
...
   const [isAddStaffModalOpen, setIsAddStaffModalOpen] = useState(false);
   const [editingStaff, setEditingStaff] = useState<Staff | null>(null);
+  const [filters, setFilters] = useState<Filters>({
+    facilities: [],
+    entitlements: [],
+  });
+
+  const [filteredStaff, setFilteredStaff] = useState<Staff[]>();
+  useEffect(() => {
+    setFilteredStaff(
```
cainlevy, line 68: looks like a `const filteredStaff = useMemo(...)`

src/utils/prettyLabel.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  [FacilityService.Call]: "call",
+  [FacilityService.Emessaging]: "messaging",
+  [FacilityService.InPersonContact]: "in person (with contact)",
+  [FacilityService.InPersonNoContact]: "in person (no contact)",
+  [FacilityService.VoiceCall]: "voice call",
+};
+
+export const prettyLabel = (
+  thing: Entitlement | FacilityService | string,
```
cainlevy, line 21: presumably this works as long as our enums don't collide?

cainlevy (approved):

src/components/System/SystemStaff.tsx
cainlevy: oh you commented about this while i was reviewing. hmm. i don't have any immediate ideas.

src/utils/prettyLabel.ts
wittejm: Yep. If that becomes an issue, we can rethink it; I considered it a low-effort refactor

PR #426, opened 2022-09-16
StaffTable is used in both facility and system staff pages

wittejm:
I want to add filter by name to this before I merge so that we don't lose that functionality on the facility staff page. 

Decision I made here was to query for all system staff even on the Facility Staff page, and then filter locally. This makes the code a lot simpler, also has some savings in the cache due to the matching data structure; querying by facility would mean fetching `facility.staffPositions` and needing to reorganized that into an array of `staff`. 

https://user-images.githubusercontent.com/2104990/190816219-f3bbfe4a-0f77-4c0c-a30e-d1a563001cc8.mov


cainlevy (approved):

wittejm:
Added the text field and ready for another review @cainlevy 

https://user-images.githubusercontent.com/2104990/191369226-e58b0df5-a41f-4405-acfb-9e046f3cb31d.mov

src/components/SelectableFieldStringFilter.tsx
```diff
@@ -0,0 +1,46 @@
...
+  onSelectChange,
+  onTextChange,
+  filterToLabelMap,
+  selected,
+  text,
+  selectStyle,
+  searchStyle,
+}: Props) => {
+  return (
+    <Input.Group key="filter" compact>
```
cainlevy, line 23: `key` is probably not effective here

src/pages/InmatesPage.tsx
```diff
@@ -87,25 +88,13 @@ const InmatesPage: React.FC = () => {
...
-              onChange={(event) => setSearchQuery(event.target.value)}
-            />
-          </Input.Group>,
+          <SelectableFieldStringFilter
+            onSelectChange={(value) => setActiveSearchFilter(value as TFilter)}
+            onTextChange={(value) => setSearchQuery(value)}
+            selected={activeSearchFilter}
+            text={searchQuery}
+            filterToLabelMap={FILTER_TO_LABEL_MAP}
+          />,
```
cainlevy, line 97: `key` goes here now

src/pages/VisitorsPage/Visitors.tsx
```diff
@@ -56,25 +57,13 @@ const Visitors = ({ visitors, loading }: Props) => {
...
-              onChange={(event) => setSearchQuery(event.target.value)}
-            />
-          </Input.Group>,
+          <SelectableFieldStringFilter
+            onSelectChange={(value) => setActiveSearchFilter(value as TFilter)}
+            onTextChange={(value) => setSearchQuery(value)}
+            selected={activeSearchFilter}
+            text={searchQuery}
+            filterToLabelMap={FILTER_TO_LABEL_MAP}
+          />,
```
cainlevy, line 66: `key` goes here now

cainlevy (approved):

PR #427, opened 2022-09-16
upgrade to react-scripts v5

cainlevy:
Upgrading to react-scripts v5 fixes a large number of listed vulnerable dependencies. It's not a light exercise, though, because we have to drop craco and work through some webpack 5 issues.

The only remaining peer dependency conflict was `graphql-request` from our codegen tools, which was an easy bump to sneak in.

I verified this release by logging in and clicking around the app. I didn't see any styling issues or new errors.


package.json
```diff
@@ -85,16 +90,17 @@
...
     "@types/socket.io-client": "^1.4.34",
-    "@typescript-eslint/eslint-plugin": "^5.3.0",
-    "@typescript-eslint/parser": "^5.3.0",
-    "autoprefixer": "^9.8.6",
-    "eslint-config-airbnb": "^18.2.1",
+    "@typescript-eslint/eslint-plugin": "^5.23.0",
+    "@typescript-eslint/parser": "^5.23.0",
+    "eslint": "8.15.0",
+    "eslint-config-airbnb": "^19.0.4",
+    "eslint-config-react-app": "^7.0.1",
```
cainlevy, line 97: dealing with eslint properly is a whole other branch. i was trying to resolve some weird behavior on startup. can't be sure if all of these changes were necessary, but i figure they're not harmful.

package.json
```diff
@@ -85,16 +90,17 @@
...
     "husky": "^7.0.4",
     "lint-staged": "^11.2.6",
-    "postcss": "^7.0.35",
+    "postcss-flexbugs-fixes": "^5.0.2",
+    "postcss-preset-env": "^7.8.2",
```
cainlevy, line 101: the new tailwindcss and postcss support doesn't include these plugins, which need to exist even though they're not mentioned or referenced by our code.

package.json
```diff
@@ -85,16 +90,17 @@
...
     "prettier": "^2.4.1",
     "prettier-plugin-organize-imports": "^2.3.4",
-    "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.1.2",
-    "typescript": "^4.4.4"
+    "typescript": "^4.8.3"
```
cainlevy, line 104: typescript 4.5+ was necessary to fix an import path from csv-parse that became a problem with webpack 5

postcss.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  plugins: {
+    tailwindcss: {},
+    autoprefixer: {},
+  },
+}
```
cainlevy, line 6: must exist when tailwind.config.js exists

package.json
cainlevy: nope! gotta do it now: https://github.com/Ameelio/connect-doc-client/actions/runs/3071124371/jobs/4961534579

package.json
```diff
@@ -85,16 +89,18 @@
...
     "husky": "^7.0.4",
     "lint-staged": "^11.2.6",
-    "postcss": "^7.0.35",
+    "postcss-flexbugs-fixes": "^5.0.2",
+    "postcss-preset-env": "^7.8.2",
     "prettier": "^2.4.1",
     "prettier-plugin-organize-imports": "^2.3.4",
-    "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.1.2",
-    "typescript": "^4.4.4"
+    "react-scripts": "5.0.1",
```
cainlevy, line 103: moving this to dev dependencies means we can successfully audit against only production advisories

src/components/Settings/EditScheduleForm.tsx
```diff
@@ -49,7 +49,7 @@ const EditScheduleForm: React.FC<Props> = ({
       visitorTypes: schedule.visitorTypes,
       meetingSlots: scheduleToWeeklySchedule(schedule.meetingSlots),
     });
-  }, [schedule]);
+  }, [form, schedule]);
```
jackofsometrade, line 52: IDK what `setFieldsValue` does, it it possible for this to create a useEffect loop?

cainlevy: will verify!

cainlevy: this is somewhat counter intuitive but it works. i verified that the `useEffect` runs exactly once when loading the form, and that the form values were populated correctly.

jackofsometrade (approved):

PR #428, opened 2022-09-19
adopt react-router-dom@6

cainlevy:
Small update to react-router-dom@6 with a familiar `useRoutes` setup.

This also trades out the HashRouter for a BrowserRouter. Some time ago we relied on hash navigation because the app was hosted on GitHub pages. Now that we deploy our own static site server with nginx, we can use standard paths.


deploy/nginx.conf
```diff
@@ -91,7 +91,7 @@ http {
         client_max_body_size 6M;
 
         location / {
-
+            try_files $uri $uri/ /index.html;
```
cainlevy, line 94: copied this over from providers so that all unknown routes will be handled by react

src/components/ErrorBoundary.tsx
```diff
@@ -18,7 +15,7 @@ const ErrorBoundary: React.FC = ({ children }) => {
               <Button
                 key="reset"
                 onClick={() => {
-                  history.goBack();
+                  window.history.back();
```
cainlevy, line 18: error boundary doesn't have router context

src/components/hocs/ProtectedRoute.tsx
```diff
@@ -1,22 +1,15 @@
...
 
-export interface ProtectedRouteProps extends RouteProps {
-  authenticationPath: string;
-}
-
-export const ProtectedRoute: React.FC<ProtectedRouteProps> = ({
-  authenticationPath,
-  ...routeProps
-}) => {
+const ProtectedRoute = () => {
```
cainlevy, line 5: this becomes analogous to SessionBoundary and may get upgraded some day

wittejm (approved):

PR #432, opened 2022-09-21
explicitly import Buffer type

wittejm:
Solution found here:
https://stackoverflow.com/questions/67804053/uncaught-referenceerror-buffer-is-not-defined-in-react

There seem to be other fresh bugs in the csv parser, probably since the version bump from 5 days ago. But this is one required fix and I did get a successful validation/upload to work locally.


cainlevy (approved):
webpack 5 stopped including nodejs polyfills automatically. i thought i'd fixed this by importing a different `csv-parse` path. this looks necessary though!

PR #433, opened 2022-09-21
fix path syntax for requests pages

cainlevy:


jackofsometrade (approved):
rubber stamp for time

cainlevy (commented):
i searched the application for mentions of `connection-requests` and `call-requests` and could not see any cases where the `:type` segment would be missing

FreedomBen (approved):

PR #434, opened 2022-09-22
Edit Staff cache updates

crcommons:
In investigating a completely different bug, I noticed that cache wasn't updating here. I attempted cache modification directly, but because the data is VERY nested, it was becoming a nightmare. Decided refetch was the best choice.


cainlevy (approved):
what data was not updating?

crcommons:
> what data was not updating?

The nested data: facility and entitlements

PR #435, opened 2022-09-22
disable closing drawer while editing schedule

cainlevy:


crcommons (approved):
haha I was going some long route where if you clicked out of the drawer, it would save your changes so they'd still be there when you reopened, but it also had to check to see if you were working on the same form as before so it could clear the changes if not. This is a far superior solution

PR #437, opened 2022-09-23
Jw/sc 4630/more on staff detail page

wittejm:
version bump is in commit 2. PRing now for sanity check on this antd version bump. I need `4.17.0` to support this snazzy [Multiple Cascader component](https://ant.design/components/cascader/) that @eliz214 requested. Decided to bump to the latest. When I bumped the version I got two type errors and fixed them here. I also clicked around the app and didn't see any issues. Anything else I should consider about bumping this version?


wittejm:
And commit 1 adds entitlements and facilities to detail page like so: 
Screen Shot 2022-09-23 at 12 10 45 PM

cainlevy (approved):
LGTM. i'm guessing that antd is pretty good about semver.

PR #438, opened 2022-09-28
Jw/sc 4630/staff timeline hifi

wittejm:
https://user-images.githubusercontent.com/2104990/192903829-81213541-d8c8-493c-873d-5047a54e99da.mov


src/components/Staff/StaffTimeline.tsx
```diff
@@ -94,20 +74,82 @@ const StaffTimeline = ({ staffId, facilityId, systemId }: Props) => {
...
   const [filters, setFilters] = useState<Filters>({
     facilities: facilityId ? [facilityId] : [],
+    eventTypes: [],
   });
+
+  const [selections, setSelections] = useState<string[][]>(
+    facilityId ? [["facilities", facilityId]] : []
+  );
+
+  useEffect(() => {
```
cainlevy, line 112: useMemo? i'd be happy to pair on the issue if there's still some unknown behavior with dependencies. those dependencies really shouldn't be different on `useEffect` vs `useMemo`.

src/components/Staff/StaffTimeline.tsx
```diff
@@ -94,20 +74,82 @@ const StaffTimeline = ({ staffId, facilityId, systemId }: Props) => {
...
+    // if a parent (All facilities or All eventTypes) is selected, it appears as ['facilities'] or ['eventTypes']
+    // if some but not all children of a parent are selected, they each appear as ['facilities', facilityId]
+    setFilters({
+      facilities:
+        selections.some(
+          (selection) => selection.length === 1 && selection[0] === "facilities"
+        ) && data
+          ? data?.system.facilities.map((f) => f.id) // If the parent is selected, include all facility ids
```
cainlevy, line 120: the `setFilteredTimeline` logic below appears to understand `filters.facilities.length === 0` as equivalent to this logic. would an empty array be simpler here?

if so, then i think there's also a possibility that the ternary might be collapsed into something like:

```
selections.filter(isFacilitySelection).map(getFacilityId).filter(notUndefined)
```

src/components/Staff/StaffTimeline.tsx
```diff
@@ -94,20 +74,82 @@ const StaffTimeline = ({ staffId, facilityId, systemId }: Props) => {
...
+          : selections
+              .filter(
+                (entry) => entry.length === 2 && entry[0] === "facilities"
+              )
+              .map((entry) => entry[1]),
+      eventTypes: selections.some(
+        (selection) => selection.length === 1 && selection[0] === "eventTypes"
+      )
+        ? filterSelectOptions[1].children?.map((c) => c.value as string) || [] // If the parent is selected, include all eventTypes
```
cainlevy, line 129: would this also work given the `setFilteredTimeline` logic below?
```suggestion
        ? []
```

src/utils/entitlementLabel.ts
```diff
@@ -0,0 +1,16 @@
+import { Entitlement } from "src/api/graphql";
+
+const entitlementLabel = (entitlement: Entitlement | string) => {
```
cainlevy, line 3: why accept arbitrary strings? are there situations where we call this function without knowing if we have an Entitlement?

cainlevy (approved):

src/utils/entitlementLabel.ts
wittejm: There was one, but I see a way to rework it and remove this (which is admittedly weird)

src/components/Staff/StaffTimeline.tsx
wittejm: made a ticket. pairing soon sounds good! https://app.shortcut.com/ameelio/story/4688/replace-useeffect-usestate-with-usememo-in-staff-pages

PR #444, opened 2022-09-30
show 'auto-approved' on call in search calls table if there was no approver

wittejm:


src/components/SearchMeetings/SearchMeetingsTable.tsx
```diff
@@ -128,9 +128,13 @@ const SearchMeetingsTable: React.FC<Props> = ({
           title="Reviewed By"
           dataIndex="reviewedBy"
           key="reviewedBy"
-          render={(staffPosition: StaffPosition) => (
+          render={(staffPosition?: StaffPosition) => (
```
cainlevy, line 131: ah! here's a pattern that i think we can adopt to improve our type safety in these antd tables:

```suggestion
          render={(staffPosition: MyMeeting['reviewedBy']) => (
```

cainlevy (approved):

PR #446, opened 2022-10-07
New RequestRecordingButton

dabbler0:


src/graphql/GetFacilityMeetings.graphql
```diff
@@ -39,6 +39,18 @@ query GetFacilityMeetings(
             enabled
             meetingType
           }
+          call {
+            ...CallFragment
+            recording {
```
dabbler0, line 44: This is necessary for including the "Request Recording" button in the search table, and may or may not cause a performance hit depending on how well our DataLoaders do batching. Open question as to whether we want to launch with this.

src/api/client.ts
```diff
@@ -59,6 +59,7 @@ export const createApolloClient = (
       authLink,
       httpLink,
     ]),
+    connectToDevTools: true,
```
cainlevy, line 62: what's the use case for dev tools in production?

src/components/RequestRecordingButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+    update: (cache, { data }) => {
+      cache.modify({
+        id: cache.identify({ __typename: "Call", id: meeting.call?.id }),
+        fields: {
+          recording: () => ({
+            status: RecordingStatus.Pending,
+          }),
+        },
+      });
+    },
```
cainlevy, line 44: i saw that your recent API change updates the recording status during this mutation. maybe we could fetch that data back from `CreateRecording.graphql` to avoid this manual cache manipulation?

src/components/RequestRecordingButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const now = new Date().getTime();
+
+  if (
+    !meeting.call ||
+    [MeetingStatus.NoShow, MeetingStatus.Cancelled].includes(meeting.status)
+  ) {
+    disabledReason = "This meeting never took place.";
+    enabled = false;
+  }
+  if (meeting.call!.expiresAt! <= now) {
```
cainlevy, line 73: this looks like it still has to worry about the `meeting.call == null` scenario. maybe these guards could use early returns to progressively eliminate their edge cases:

```
if (something) {
  return <DisabledButton reason={...} />;
}
```

src/graphql/GetFacilityMeetings.graphql
cainlevy: yeah, hmm, i think the data loaders should handle this alright. but i'm still unable to predict when we'll see the kinds of degenerate `await` behavior that caused issues the other month. thankfully this only affects a single page and not the whole login process.

let's keep a close eye on this when it goes out. we can test some live queries and roll back if there are any issues.

cainlevy (approved):

PR #448, opened 2022-10-08
Tywarren/sc 4625/functional recording page

jackofsometrade:
https://user-images.githubusercontent.com/97806603/194677518-66ccb032-299b-4c54-ae04-72ae11be54b2.mp4


src/components/SearchMeetings/SearchMeetingsTable.tsx
```diff
@@ -182,22 +185,34 @@ const SearchMeetingsTableActionButton: React.FC<{
...
+        {(meeting.status === MeetingStatus.Scheduled ||
+          meeting.status === MeetingStatus.Live) && (
+          <CancelMeetingButton meeting={meeting} />
+        )}
+        {(meeting.status === MeetingStatus.Ended ||
+          meeting.status === MeetingStatus.Terminated) &&
         meeting.meetingType === MeetingType.Call &&
-        !meeting.confidential && (
+        !meeting.confidential &&
+        meeting.call?.recording?.status === RecordingStatus.Available ? (
```
cainlevy, line 205: this reminds me that anthony has some work in the area you may want to be aware of: https://github.com/Ameelio/connect-doc-client/pull/446/files

dabbler0: Merged these in PR #450

src/pages/RecordingPage.tsx
```diff
@@ -0,0 +1,240 @@
...
+          <Layout.Content>
+            <Row gutter={16}>
+              <Col span={8}>
+                <Descriptions
+                  size="small"
+                  column={1}
+                  bordered
+                  style={{ backgroundColor: "white" }}
+                >
+                  <Descriptions.Item label="Recording Link">
```
dabbler0, line 128: Do we need to ever expose the raw recording link other than through the "download" button? It seems like we should encourage sharing the `/recording/{meeting.id}` link instead. This way it sits behind authorization and also doesn't expire until the artifacts are actually gone. Then this row of the table can be used for the actual artifact expiration date instead of the (short) link expiration date.

dabbler0: (We can sync with design on this question; approving for now)

dabbler0 (approved):

src/components/Recordings/RecordingPlaceholder.tsx
```diff
@@ -0,0 +1,45 @@
+import { PlayCircleFilled } from "@ant-design/icons";
+import React from "react";
+import { generateBgColor } from "src/utils";
+import { InitialsAvatar } from "../Avatar/InitialsAvatar";
+import "./RecordingPlaceholder.css";
```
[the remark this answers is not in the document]
jackofsometrade: I can't actually.  I saw this pattern being used in other components and continued the trend.  If there's some lost benefits using this method I can switch to antd+tailwind.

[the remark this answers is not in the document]
cainlevy: ahh, yeah, most of the css files i've seen in the doc client exist because antd is otherwise resistant to customization. or in some cases they just appear to be very old.

tailwind can be a bit of a headache but it currently handles our spacing and color values.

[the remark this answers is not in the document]
jackofsometrade: I took another look at this and the RecordingPlacerholder is really a custom design (thanks @eliz214) that behaves more like a series of images than html elements.  It gives us a flexible row of N participants with the play button rendered on top.  I think custom CSS is the better solution here, as the contents of the placeholder don't relate very much to tailwind or our recurring styles.

dabbler0 (approved):

src/components/RequestRecordingButton.tsx
```diff
@@ -12,6 +12,7 @@ import { CreateRecordingDocument } from "src/graphql/CreateRecording.generated";
 import { errorReporter } from "src/utils";
 import { useMutation } from "@apollo/client";
 import { useAppSelector } from "src/redux";
+import { useNavigate } from "react-router-dom";
```
jackofsometrade, line 15: @dabbler0 here 👉

src/components/RequestRecordingButton.tsx
```diff
@@ -51,7 +53,12 @@ const RequestRecordingButton: React.FC<Props> = ({ meeting, ...rest }) => {
 
   if (ready) {
     return (
-      <Button href={meeting.call!.recording!.url!} {...rest}>
+      <Button
```
jackofsometrade, line 56: @dabbler0 here 👉

jackofsometrade:
@dabbler0 re-requesting review due to a couple small changes 🙏 commented on 2 and the rest is graphql generated query updates.

PR #449, opened 2022-10-11
refactor with SessionBoundary

cainlevy:
This is an overhaul of the login and session system following patterns established in the providers app:

- **Routes declare layouts** — If you don't want the menu, simply choose a different layout.
- **SessionBoundary loads the current user** — LoginPage is only responsible for getting the session token.
- **Current user information is read directly from the Apollo cache** — Redux gets a lot smaller.


src/App.tsx
```diff
@@ -35,71 +20,23 @@ if (process.env.NODE_ENV === "production") {
...
-  }, [session.status, session.user, dispatch]);
-
-  // redirect to login
-  useEffect(() => {
-    if (!isAuthenticated && pathname !== LOGIN_PATH)
-      dispatch(setRedirectUrl(pathname));
-  }, [dispatch, isAuthenticated, pathname]);
+  const isLoggedIn = !!session.token;
 
   // on logout, clear apollo cache
```
cainlevy, line 29: this is the last bit of interesting logic in App.tsx, and it will eventually move to an AuthProvider

src/components/LiveCall/CallRegistry.tsx
```diff
@@ -42,8 +44,15 @@ export default function CallRegistry({ calls, children }: Props) {
...
   return (
     <CallRegistryContext.Provider value={registry}>
-      {calls.map((c) => (
-        <RegisterCall key={c.id} call={c} register={register} drop={drop} />
+      {liveCalls.map((c) => (
+        <RegisterCall
+          key={c.id}
+          callId={c.id}
+          callToken={c.call?.token!}
+          callUrl={c.call?.url!}
```
cainlevy, line 52: destructuring `call` into individual props allows `React.memo()` to work better so that `CallRegistry` can fetch its own list of live calls

src/components/Menu/Menu.tsx
```diff
@@ -164,8 +161,6 @@ export default function Menu({
       ? personalConnections.length + pendingPersonalMeetings.length
       : 0);
 
-  if (!isLoggedIn) return <div />;
```
cainlevy, line 167: no longer a worry thanks to declarative layouts

src/index.tsx
```diff
@@ -1,25 +1,17 @@
...
   <React.StrictMode>
     <Provider store={Store}>
-      <ApolloProvider client={createApolloClient(Store, showToast)}>
-        <ErrorBoundary>
-          <BrowserRouter>
-            <App />
-          </BrowserRouter>
-        </ErrorBoundary>
-      </ApolloProvider>
+      <BrowserRouter>
```
cainlevy, line 12: i wanted to move more into App.tsx but this is necessary to enable `useRoutes` on the next level down

src/redux/modules/session.ts
```diff
@@ -1,88 +1,44 @@
...
-
-type SessionStatus = "inactive" | "active" | "loading";
 
 interface SessionState {
-  status: SessionStatus;
-  user: User;
   token: string;
-  redirectUrl: string;
-  authInfo: AuthInfo;
   selected?: SelectedFacility;
```
cainlevy, line 5: i'd love to put this responsibility in `FacilityLayout` but we still have references from system pages. haven't decided whether to address that debt or compromise to finish off redux.

wittejm (approved):

PR #451, opened 2022-10-12
Expose expiration date of recording artifacts

dabbler0:


src/pages/MeetingDetailsPage.tsx
```diff
@@ -96,6 +106,15 @@ function MeetingDetailsPage(): ReactElement {
             <Descriptions.Item label="Meeting ID">
               {meeting.id}
             </Descriptions.Item>
+            <Descriptions.Item label="Recording expires">
+              <Text type="danger">
+                {meeting.call &&
+                  `${formatDate(
+                    meeting.call!.expiresAt!,
+                    "dateyear"
+                  )} at ${formatDate(meeting.call!.expiresAt!, "time")}`}
```
cainlevy, line 115: looks like we can remove a couple of `!` because of the `meeting.call &&` guard condition:

```suggestion
                  `${formatDate(
                    meeting.call.expiresAt!,
                    "dateyear"
                  )} at ${formatDate(meeting.call.expiresAt!, "time")}`}
```

(at a previous gig i learned that my `!` overrides turned into bugs more often than not so i try to minimize them)

cainlevy (approved):

PR #454, opened 2022-10-12
make background white, not transparent

enrjp:
Very simple PR to fix the issue with the background: transparent rule which is default applied to the antd icons. The problem in this case is that the element underneath the icon has a dark background, so the dots blend in with it. I checked ant to see if there was any theme property to be applied to buttons and icons, like dark or night, to switch the primary and background colors, but I did not see anything. If you know of something I missed do let me know?

The simple addition looks like the following:

Default:
Screen Shot 2022-10-12 at 6 50 22 PM

On hover:
Screen Shot 2022-10-12 at 6 50 25 PM

I don't think the styling looks _great_ but it is much more usable at least.


cainlevy (approved):
oh maybe this broke in an antd dependency update. i think we had one a month or two ago.

thanks for fixing!

PR #461, opened 2022-10-18
fix a hard-coded 7 days and remove redundant expiration

cainlevy:


src/pages/RecordingPage.tsx
```diff
@@ -125,7 +125,7 @@ export default function RecordingPage() {
         <Result
           status="404"
           title="Expired link"
-          subTitle="All recordings permanently expire 7 days after the meeting start time unless previously downloaded."
+          subTitle={`All recordings permanently expire ${meeting.facility.system.recordingRetention} days after the meeting start time unless previously downloaded.`}
```
dabbler0, line 128: In the course of implementing some other features I also loaded the facility system in a way that's available through `useSelectedFacility`. Might use that instead of expanding the gql request?

dabbler0 (approved):

PR #462, opened 2022-10-18
Fetch id when fetching system in `ImportDetails` to avoid cache overwrite

dabbler0:
Roster validation was effectively causing "logout" because:
 1. To be logged in you need certain details on the current `facility.system`
 2. The graphql query to validate a roster fetched some very sparse details on `facility.system`, and did *not* fetch `facility.system.id`
 3. So the `facility.system` property was overwritten with a plain object instead of maintaining a reference to the cached `System` object
 4. So `facilty.system` retained none of the previous cached `System` properties, which now appear missing, so the client thinks it is logged out.

The fix is to fetch `id` when fetching the system in `GetImportDetails`. Hopefully this was the only problem.

I'm thinking we should probably have a policy of always fetching `id` on all entities we ever request through Apollo; this is not the first time the Apollo cache has behaved erratically because of a missing `id` in a request.


cainlevy (approved):
oooo yes we absolutely want to include `id` everywhere appropriate.

i also don't know how to enforce that.

jackofsometrade:
> i also don't know how to enforce that.

First time looking but could we use something like this?  https://github.com/B2o5T/graphql-eslint/blob/master/docs/rules/require-id-when-available.md

cainlevy:
Wow! Maybe? How does it even ...

enrjp:
@dabbler0 @cainlevy @jackofsometrade Interestingly I'm wondering if this may be related to behavior I've seen in the DOC Client when developing locally. After logging in, if you refresh the page, I get logged out. I'd like to log a bug ticket in Shortcut for me to investigate, but first I should ask if anyone else has experienced this same behavior, here's a quick screen recording:

https://user-images.githubusercontent.com/5834080/196573705-bcf3c12c-b231-4397-8691-e807a50bdb1d.mov

In this recording I navigate to a few pages using the React Router, but you'll see when I refresh the page with Command + R I get re-directed back to the Login page...

PR #463, opened 2022-10-18
Sc 2024

enrjp:
This PR adds validation to the form for creating/editing a contact restriction. With the inclusion of the Form.Item component, we now can show an error message if the textarea previously had input but becomes empty. Example is seen below:

Screen Shot 2022-10-18 at 4 22 57 PM

Once the comment is provided, as one would expect, the error messaging and shading disappears:
Screen Shot 2022-10-18 at 4 23 03 PM


src/components/ConnectionDetailDrawer/EditPermissionsModal.tsx
```diff
@@ -189,11 +192,21 @@ const EditPermissionsModal = ({
...
+                rules={[
+                  {
+                    required: true,
+                    message: "A comment is required",
+                  },
+                ]}
+              >
+                <Input.TextArea
+                  defaultValue={restriction ? restriction.reason : ""}
+                  onChange={(e) => setReason(e.target.value.trim())}
```
enrjp, line 207: I added the invocation to trim() as well. This prevents a user from entering a bunch of spaces to trick the validation and also tightly formats their input in case they prefixed or suffixed their input with an errant space character.

src/components/ConnectionDetailDrawer/EditPermissionsModal.tsx
```diff
@@ -147,9 +147,12 @@ const EditPermissionsModal = ({
       >
         <Row className="flex justify-between">
           <Row>
-            <Typography.Text>{`${inmate.firstName} ${inmate.lastName} (${inmate.inmateIdentification})`}</Typography.Text>
-            {` & `}
-            <Typography.Text>{`${contact.firstName} ${contact.lastName}`}</Typography.Text>
+            <Typography.Text>
+              {`${inmate.firstName} ${inmate.lastName} (${inmate.inmateIdentification})`}{" "}
+              & {`${contact.firstName} ${contact.lastName}`}
+            </Typography.Text>
```
enrjp, line 153: This change is to fix this little visual issue I noticed:
Screen Shot 2022-10-18 at 3 40 47 PM

As you can see the name of the first participant in the connection is missing a space between the last letter and the & character. So I fixed it, as seen below:

Screen Shot 2022-10-18 at 3 40 57 PM

src/components/ConnectionDetailDrawer/EditPermissionsModal.tsx
```diff
@@ -147,9 +147,12 @@ const EditPermissionsModal = ({
...
+
+            <Typography.Text></Typography.Text>
```
wittejm, line 155: is this element also part of the fix?

enrjp: Thank you for catching that!! I think it was accidentally added as an autocomplete snippet... I've now removed 👍🏻

wittejm (approved):

PR #466, opened 2022-10-21
JR/sc-2505/add group name to inmate profile page

enrjp:
Simple PR to update the Inmate Profile page with the 'Group' they belong to.

Example seen below:
Screen Shot 2022-10-21 at 11 03 05 AM


src/pages/InmatePage.tsx
```diff
@@ -94,6 +94,9 @@ function InmatePage(): ReactElement {
           <Descriptions.Item label="Needs Call Approval">
             {inmate.needsApproval ? "Yes" : "No"}
           </Descriptions.Item>
+          <Descriptions.Item label="Group">
+            {inmate.group.name}
+          </Descriptions.Item>
```
enrjp, line 99: Just adding another Descriptions.Item component instance, simple 👍🏻

src/graphql/fragments/InmateFragment.graphql
```diff
@@ -11,4 +11,8 @@ fragment InmateFragment on Inmate {
   malanId
   needsApproval
   needsMonitor
+  group {
+    id
+    name
+  }
```
enrjp, line 17: We needed to request the group in the InmateFragment, but as you can see, this updates A LOT of other queries that utilize this fragment, and they have no need for the group yet, so this is unnecessary retrieval overhead from the DB (if not cached). @dabbler0 or @cainlevy I noticed that there are a lot of fragments that request fields that not every page that utilizes the query needs, I might imagine this is mitigated through caching?

cainlevy, line 17: sorry jumping in here to say that our baseline policy is that fragments should only load simple (cheap) fields from the immediate record. this makes them composable with fewer side effects on query plans.

cainlevy: good observation. we want fragments to be composable with minimal side effects on query planning so we would normally move this `group { ... }` loading to the relevant query. e.g.:

```
inmate {
  ...InmateFragment
  group {
    ...GroupFragment
  }
}
```

enrjp: Gotcha! Thanks for explaining where to put it...

src/graphql/GetInmate.graphql
```diff
@@ -1,6 +1,9 @@
 query GetInmate($inmateId: ID!) {
   inmate(id: $inmateId) {
     ...InmateFragment
```
enrjp, line 3: @cainlevy rather than utilizing the whole ...GroupFragment, I just needed the group's name, so I added below...

cainlevy: that's fine. especially since it looks like GroupFragment itself doesn't apply the convention we're following here!

you'll want to always query for `id` when available, though, so that apollo can properly normalize and merge results from different queries

cainlevy (approved):

src/graphql/GetInmate.graphql
enrjp: oops sorry, yeah I should've added that - PR updated, will merge 👍🏻

PR #475, opened 2022-12-02
JR/SC-2266/emessaging use rejection reason codes instead

enrjp:


src/api/graphql.ts
```diff
@@ -620,6 +620,7 @@ export type EndAllVoiceCallsResponse = {
```
enrjp, line 620: This is just from running update graphql 👍🏻

src/components/MessageReviewPanel/MessageReviewControlPanel.tsx
```diff
@@ -8,11 +9,6 @@ interface Props {
   loading?: boolean;
 }
 
-const REJECTION_OPTIONS = [
-  "violates the terms of service",
-  "contains prohibited language",
-];
```
enrjp, line 14: Rejection options were extracted into an enum in constants as seen in the import

src/constants/Call.ts
```diff
@@ -1,28 +1,35 @@
 import { RejectionReason } from "src/typings/Common";
 
+enum CallRejectionReason {
+  Storm = "storm",
+  Lockdown = "lockdown",
+  DeviceUnavailable = "broken",
+  Sickness = "sick",
+}
```
enrjp, line 8: Created a new enum to represent call rejection reasons

src/constants/Call.ts
```diff
@@ -1,28 +1,35 @@
...
+
 export const DEFAULT_CALL_REJECTION_REASONS: RejectionReason[] = [
   {
+    key: CallRejectionReason.Storm,
```
enrjp, line 12: And replaced the string key with the enum key

src/constants/Connection.ts
```diff
@@ -0,0 +1,28 @@
+import { RejectionReason } from "src/typings/Common";
```
enrjp, line 1: Added a new file to constants to represent connection rejection reasons

src/constants/Message.ts
```diff
@@ -0,0 +1,19 @@
+import { RejectionReason } from "src/typings/Common";
```
enrjp, line 1: Added a new file to constants to represent message rejection reasons

src/pages/ConnectionRequestsPage/ConnectionReviewModal.tsx
```diff
@@ -5,18 +5,16 @@ import Text from "antd/lib/typography/Text";
...
+import { RejectionReason } from "src/typings/Common";
 import { genFullName } from "src/utils";
 import IdentityStatusTag from "../../components/IdentityStatusTag";
 import InmateCell from "../../components/Requests/InmateCell";
 import "./ReviewModal.css";
 
-export type ConnectionRejectionReason = {
-  title: string;
-  description: string;
-};
```
enrjp, line 18: Can re-use the RejectionReason type, don't really need this one

cainlevy (commented):
Sorry I think there was some misunderstanding in this ticket. What I was proposing was that we send a code (e.g. "storm" or "lockdown") to the API and avoid converting it into user-facing copy until the last moment. That would let us translate the reason into a user's preferred language.

Translation is likely to become more important for us. Now that I've been able to spend more time to learn about product priorities, however, I think it's most likely that the feature we actually need is the ability for facilities to customize their set of available options. That may end up with a very different implementation so we may want to hold off.

enrjp:
> Sorry I think there was some misunderstanding in this ticket. What I was proposing was that we send a code (e.g. "storm" or "lockdown") to the API and avoid converting it into user-facing copy until the last moment. That would let us translate the reason into a user's preferred language.
> 
> Translation is likely to become more important for us. Now that I've been able to spend more time to learn about product priorities, however, I think it's most likely that the feature we actually need is the ability for facilities to customize their set of available options. That may end up with a very different implementation so we may want to hold off.

Ahhhh ok, I understand now 👍🏻 . Should I still implement these API codes in the meantime? Or even if not, I feel like this PR is a bit of clean-up/re-factor of what was previously there so might be good to merge anyways?

cainlevy:
Let's hold off on API changes. I haven't studied this PR yet -- would you make any changes to it knowing that it's purely a cleanup / refactor and not required to complete any current tickets? Or is it ready for review as-is?

enrjp:
> Let's hold off on API changes. I haven't studied this PR yet -- would you make any changes to it knowing that it's purely a cleanup / refactor and not required to complete any current tickets? Or is it ready for review as-is?

Yes this is just a cleanup / refactor in its current form. It doesn't enable (or block) any other tickets in the backlog, at this stage it would just be a standalone for the DOC client. No rush or anything to review/merge, I can leave in a draft state if that's preferred 👍🏻

cainlevy (approved):

PR #477, opened 2022-12-02
JR/SC-5111/forbiddenerror-getstaff

enrjp:
This bug was observed when attempting to retrieve a StaffPosition record via GetStaff. I checked the user making the request and it looks like it's a user in the Iowa system. The user whose staffPositions they are attempting to retrieve is the Connect Support user. The Connect Support user has staff position records across multiple facilities, many of which I would presume are not within the Iowa system. So I believe because the staff positions being requested aren't filtered by just the requesting user's own system ID, the policy rule becomes violated when a position record is retrieved associated with a facility that is outside their system.

So in this PR I've added an optional param systemId that is passed to the staffPositions field. I believe this should be sufficient as I already saw we have support on the backend for filtering staffPositions by systemId. If the positions fetched are associated with facilities having a system ID that match the requesting user's system ID, the policy rule function should return true.


cainlevy (approved):
💯

